PR #3, opened 2017-10-05
Fix incorrect logo URL

pluehne:
Unfortunately, I referenced the logo with an absolute URL, which doesn’t work on the hosted page at [autodesk.github.io/hubble](). Also, I accidentally included the logo twice. This pull requests corrects these issues to make the logo render properly.


larsxschneider:
Thank you 👍 
BTW: Do you think it would look cool to have the star background in the header?

pluehne:
I also thought about adding the starscape background to the header earlier. Let me later give a try how it looks 😃.

pluehne:
Well, it looks nice enough to me 😃. I now put in a starscape background, though I might fine-tune it at some point.

PR #4, opened 2017-10-09
Switch to MIT license

pluehne:
I propose switching Hubble to the MIT license.

The reasons are mainly that the MIT license is more concise and permissive than the current Apache 2.0 license. Furthermore, there is only little benefit for us in forcing forks of this project to list all of their modifications—most of them will do that anyway.

By the way, I would suggest removing the License section from the readme. There is a separate file for that, and GitHub already shows the used license prominently in the header bar 😃.


larsxschneider:
@craigez @toddocon @rashamalek @dmittman 😄

craigez:
👍 Thanks! 

Personally, I would have left a license reference in the README file.

larsxschneider:
> Personally, I would have left a license reference in the README file.
Just a gut feeling or do you some argument/best practice we could/should follow here?

craigez:
It's a little bit of both. I have a gut feeling that the more mention of licenses the better! Additionally, one of the default rulesets we've set up for this TODO group repo linter is to check the license file is referenced in the README: https://github.com/todogroup/repolinter/blob/master/rulesets/default.json#L8

I would also encourage adding it with the "SPDX-License-Identifier" keyword, i.e. "SPDX-License-Identifier: [MIT](LICENSE)".

craigez:
See #12

craigez:
Can you hold off on #12? I used the wrong email.

Craig

pluehne:
@craigez: No worries, feel free to rebase the branch!

PR #5, opened 2017-10-09
Make data labels consistent

pluehne:
The data labels are currently formatted inconsistently. This pull request turns all labels into lowercase (except acronyms etc.) and updates the demo data accordingly.

### DIFF
--- a/docs/demo-data/pull-request-usage.tsv
+++ b/docs/demo-data/pull-request-usage.tsv
@@ -1,4 +1,4 @@
-date	Pull request usage in %	Status usage in %
+date	pull request usage [%]	status usage [%]
 2017-09-25	88	58
 2017-09-24	88	58
 2017-09-01	85	53
--- a/updater/reports/ReportPRUsage.py
+++ b/updater/reports/ReportPRUsage.py
@@ -35,8 +35,8 @@ class ReportPRUsage(ReportDaily):
 		query = '''
 			SELECT
 				"''' + str(date) + '''" AS date,
-				ROUND(100*COUNT(pr.repository_id)/COUNT(active.repository_id)) AS "Pull request usage in %",
-				ROUND(100*COUNT(status.repository_id)/COUNT(active.repository_id)) AS "Status usage in %"
+				ROUND(100*COUNT(pr.repository_id)/COUNT(active.repository_id)) AS "pull request usage [%]",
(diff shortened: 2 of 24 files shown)


pluehne:
I forgot to update the HTML part accordingly. I will fix this!

docs/demo-data/pull-request-usage.tsv
```diff
@@ -1,4 +1,4 @@
-date	Pull request usage in %	Status usage in %
+date	pull request usage [%]	status usage [%]
```
larsxschneider: Is this some kind of standard? I think `in %` reads easier, no?

pluehne: Yep, that’s how you usually denote the unit of an axis. I think it looks more professional like this.

updater/reports/ReportPRUsage.py
```diff
@@ -35,8 +35,8 @@ def query(self, date):
 		query = '''
 			SELECT
 				"''' + str(date) + '''" AS date,
-				ROUND(100*COUNT(pr.repository_id)/COUNT(active.repository_id)) AS "Pull request usage in %",
-				ROUND(100*COUNT(status.repository_id)/COUNT(active.repository_id)) AS "Status usage in %"
+				ROUND(100*COUNT(pr.repository_id)/COUNT(active.repository_id)) AS "pull request usage [%]",
+				ROUND(100*COUNT(status.repository_id)/COUNT(active.repository_id)) AS "status usage [%]"
```
larsxschneider: are these names used anywhere? just to make sure we don't break anything here ;-)

docs/demo-data/pull-request-usage.tsv
pluehne: Also keep in mind that I want to show the unit on the axes directly at some point. Parsing might be easier with this bracket notation.

updater/reports/ReportPRUsage.py
pluehne: Yes, these are the labels shown in this chart:

screenshot from 2017-10-10 16-27-16

docs/demo-data/pull-request-usage.tsv
larsxschneider: parsing is a convincing argument! 👍

updater/reports/ReportPRUsage.py
pluehne: It’s the actual results if I see that correctly, and not just intermediate subquery results.

PR #7, opened 2017-10-10
Clarify that dataURL is a GitHub Pages URL

pluehne:
@rashamalek made me aware that the part about the `dataURL` is slightly confusing.

For this reason, I suggest adding an example URL and making it more clear that the `dataURL` is actually hosted by GitHub Pages and not just the repository URL.

@rashamalek: Do you think this is precise enough in order to set everything up correctly, or do you have another suggestion?


larsxschneider:
Looks good to me. Please merge if you are OK with my addition!

rashamalek:
@pluehne Looks good!

PR #8, opened 2017-10-10
Update GitHub audit log path for GitHub Enterprise 2.11

rashamalek:
on ghe 2.11.2 the github-audit path is a link (/var/log/github/audit.log) to /var/log/github-audit.log

this leads to an empty git-download.tsv and an empty result in hubble/housekeeping-git-traffic.

Rotated logs are kept as /var/log/github/audit.log.[1-9].*
This pr changes the ghe-audit log path in hubble/updater/script/gith-download.sh from
/var/log/github/audit.log.1*  ->  /var/log/github-audit.log.1*

### DIFF
--- a/updater/scripts/git-download.sh
+++ b/updater/scripts/git-download.sh
@@ -4,7 +4,7 @@
 #
 echo -e "repository\tuser\tcloning?\trequests/day\tdownload/day [B]"
 
-zcat -f /var/log/github/audit.log.1* |
+zcat -f /var/log/github-audit.log.1* |
     perl -ne 'print if s/.*"program":"upload-pack".*"repo_name":"([^"]+).*"user_login":"([^"]+).*"cloning":([^,]+).*"uploaded_bytes":([^ ]+).*/\1\t\2\t\3\t\4/' |
     sort |
     perl -ne '$S{$1} += $2 and $C{$1} += 1 if (/^(.+)\t(\d+)$/);END{printf("%s\t%i\t%i\n",$_,$C{$_},$S{$_}) for ( keys %S );}' |
--- a/updater/scripts/git-download.sh
+++ b/updater/scripts/git-download.sh
@@ -4,7 +4,7 @@
 #
 echo -e "repository\tuser\tcloning?\trequests/day\tdownload/day [B]"
 
-zcat -f /var/log/github-audit.log.1* |
+zcat -f /var/log/github-audit.log* |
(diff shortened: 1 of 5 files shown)


pluehne:
Thanks for this pull request, @rashamalek!

It’s interesting to hear that GitHub Enterprise changed the log file location from 2.10 to 2.11.

For now, we shouldn’t assume that users are on GitHub Enterprise ≥ 2.11 necessarily.
For this reason, it would be best to have a switch depending on which version is detected.

I won’t be able to implement the switch in the next days though, as I have many other things to do.

And for your information, I want to add a GitHub Enterprise version check at some point to make sure that Hubble runs on a supported version 😃.

larsxschneider:
@rashamalek We can use this command to check the GHE version: https://github.com/larsxschneider/scotty/blob/a68d16339a6da8cd20559ce36533f148e3376d47/lib-ghe.sh#L66-L67

We'll get this in soon as we plan to upgrade to 2.11, too 😄

updater/scripts/git-download.sh
```diff
@@ -4,7 +4,7 @@
 #
 echo -e "repository\tuser\tcloning?\trequests/day\tdownload/day [B]"
 
-zcat -f /var/log/github/audit.log.1* |
+zcat -f /var/log/github-audit.log.1* |
```
larsxschneider: Can you make this variable based on the output here: https://github.com/larsxschneider/scotty/blob/a68d16339a6da8cd20559ce36533f148e3376d47/lib-ghe.sh#L66-L67

rashamalek: Yes, Doing a version check is reasonable, will try to adapt it.

pluehne:
Thanks for the update, @rashamalek!

However, I fear hat looking for `/var/log/github-audit.log*` with a wildcard will not only give us the log files from yesterday, but all of them.
However, we currently plan to gather this data on a daily basis.

I think the best solution would be to select the file locations based on the GitHub Enterprise version in use, and to always take a `.1` log file.
What do you think?

rashamalek:
@pluehne Sorry, It was not supposed to be pushed to Autodesk/hubble, It was an internal test, will fix it

pluehne:
No worries, you can always rebase this branch 😃.

larsxschneider:
@rashamalek I cherry picked your commit and added support for pre 2.11 in #11

PR #9, opened 2017-10-11
Text review: chart descriptions

pluehne:
I just reviewed the chart description texts.

Most notably, I broke down the texts into smaller paragraphs in order to make them more legible.
I fear that people would tend not to read the texts if they are too long.

Also, I spelled out all the abbreviations, because I’m not a fan of abbreviations, especially in written text 😃.

@larsxschneider: I hope you are fine with the changes!

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -26,7 +26,7 @@
 		<script src="https://d3js.org/d3.v4.min.js"></script>
 		<script src="https://momentjs.com/downloads/moment-with-locales.min.js"></script>
 		<script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.0/Chart.min.js"></script>
-		<script src="{{ site.baseurl }}/assets/js/charts.js?version=62c28e5"></script>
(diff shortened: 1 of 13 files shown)


docs/_layouts/default.html
```diff
@@ -26,7 +26,7 @@
 		<script src="https://d3js.org/d3.v4.min.js"></script>
 		<script src="https://momentjs.com/downloads/moment-with-locales.min.js"></script>
 		<script src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.0/Chart.min.js"></script>
-		<script src="{{ site.baseurl }}/assets/js/charts.js?version=62c28e5"></script>
+		<script src="{{ site.baseurl }}/assets/js/charts.js?version=1ff0187"></script>
```
larsxschneider: is this a version update? let's not mix version updates with normal changes in the future ;-)

pluehne: No, that’s a resource tag.
You use these to force browsers to reload resources, even with browser caching enabled.

Issue #16, opened 2017-10-15
Remove inline JavaScript from header


pluehne:
The inline JavaScript tag displaying the server name in [the header bar](https://github.com/Autodesk/hubble/blob/453241289a92a473516e0bb384c7bb60f4bd18c5/docs/_layouts/default.html#L38) is [bad practice](https://robertnyman.com/2008/11/20/why-inline-css-and-javascript-code-is-such-a-bad-thing/).

I believe that we can obtain the server name statically through Jekyll instead.

PR #18, opened 2017-10-15
Remove inline JavaScript from header and replace it with Jekyll directive

GitHugop:
Ref: #16 
Uses site.url and removes the "https://" part.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name">{{ site.url | remove:'https://' | split:'/' | first }} Statistics</h1>
+				<h1 class="project-name">{{ site.url | remove:'https://' }} xStatistics</h1>
 			</a>
 		</section>
 		{% include navigation.html %}
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name">{{ site.url | remove:'https://' }} Statistics</h1>
+				<h1 class="project-name"> {{ site.url | remove:"https://" }} Statistics</h1>
 			</a>
 		</section>
 		{% include navigation.html %}
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name"><script type="text/javascript">document.write(gheHostname())</script> Statistics</h1>
+				<h1 class="project-name">{{ site.url }} Statistics</h1>
 			</a>
 		</section>
 		{% include navigation.html %}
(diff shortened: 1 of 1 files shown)


docs/_layouts/default.html
```diff
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name"><script type="text/javascript">document.write(gheHostname())</script> Statistics</h1>
+				<h1 class="project-name">{{ site.url | remove:'https://' }} Statistics</h1>
```
pluehne: Could you plase add a space after the colon and put the `https://` part in double quotes for consistency with the rest?

pluehne:
I’m not sure whether it would also be good to entirely get rid of the [`gheHostname` and `gheUrl` JavaScript functions](https://github.com/Autodesk/hubble/blob/master/docs/assets/js/charts.js#L318-L324) still used [here](https://github.com/Autodesk/hubble/blob/master/docs/assets/js/charts.js#L375).

@larsxschneider, @GitHugop: What do you guys think?

docs/_layouts/default.html
```diff
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name">{{ site.url | remove:'https://' }} Statistics</h1>
+				<h1 class="project-name"> {{ site.url | remove:"https://" }} Statistics</h1>
```
pluehne: Sorry for being ambiguous. I meant that a space should be inserted after `remove:` like this: `remove: "https://"` (and not in the beginning. Could you change that?

I know I’m a bit pedantic, but I want to keep the coding style consistent as much as possible 😃.

docs/_layouts/default.html
GitHugop: Yes, I have updated the file with your requested changes.

pluehne:
Let’s wait for @larsxschneider to review and merge this.

@larsxschneider: Do you think it would make sense to also evaluate the server name via Jekyll for the `charts.js`, as I suggested above?

GitHugop:
You welcome. Regarding the functions, I don't think it's a great idea to mix JavaScript files with Jekyll. Something that can be done is merging both functions into one, like:

`
function getURL() {
return $(location).prop("protocol") + "//" + $(location).prop("hostname");
}
`

docs/_layouts/default.html
```diff
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name"><script type="text/javascript">document.write(gheHostname())</script> Statistics</h1>
+				<h1 class="project-name">{{ site.url | remove: "https://" }} Statistics</h1>
```
larsxschneider: Thank you 👍 ! Is there a way to handle `http://` URLs, too? Some GitHub Enterprise installations use just HTTP 😢

pluehne: Wasn’t aware of this, good catch!

GitHugop: Testing it out.

docs/_layouts/default.html
```diff
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
-				<h1 class="project-name"><script type="text/javascript">document.write(gheHostname())</script> Statistics</h1>
+				<h1 class="project-name">{{ site.url | remove: "http://", "https://" }} Statistics</h1>
```
larsxschneider: I get this error with Jekyll 3.3.1 locally:
```
  Liquid Exception: Liquid error (line 38): wrong number of arguments (given 3, expected 2) in /_layouts/default.html
```
Any idea what to do about it?

GitHugop: Seems like "remove" only supports 1 argument, there doesn't seem to be a way on the Jekyll docs. We can try using both "remove" and "remove_first".

larsxschneider:
Thank you @GitHugop ! 👍 👍 👍 

@pluehne Yeah, we could use the same approach in `chart.js`.

Issue #13, opened 2017-10-13
Print the Hubble version in the header bar


larsxschneider:

austin665:
Hi @larsxschneider I will try to take a look at this, if it is ok.

larsxschneider:
That sounds awesome. Can you quickly summarize and share your plan here? E.g. the question is, how do we identify the current version? Is it possible to read with jekyll the current git hash or tag? Or should we store the value in [config.yml](https://github.com/Autodesk/hubble/blob/master/docs/_config.yml)?

austin665:
This is my plan based on a high level investigation, based on Jekyll's site configuration we can expose the github repository metadata on pages based on this https://help.github.com/articles/repository-metadata-on-github-pages/ .   I am assuming, the latest release number is the version number we want.    Please feel free to point me in the right direction, I am new to open source! 😄 

I do not have access to github enterprise, would that be an issue?

pluehne:
Just as a side note, we already have a file [`docs/updater/version`](https://github.com/Autodesk/hubble/blob/151be01d89f33fd64162c4ba371edb48cacc563f/updater/version), which contains the version number. This is currently [used for packaging](https://github.com/Autodesk/hubble/blob/151be01d89f33fd64162c4ba371edb48cacc563f/updater/packaging/debian/Makefile).

However, this might have not been the best idea—it might be better to have the version information at a single place or even just implicit, such as per looking at the latest release tag.

I just wanted to throw this in so that we don’t forget about it 😃.

PR #22, opened 2017-10-17
Cleanup GHE URL in header

larsxschneider:
If GHE subdomain isolation is enabled, then all pages URLs are 
prefixed with `pages.`. Remove this prefix in the page header. 

In addition, only remove the first occurrences of `http://` (although we 
shouldn't see that string multiple times).

@parkr is there an easier way to achieve the same? 😄 

/cc @GitHugop

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -35,7 +35,7 @@
 			</a>
 			<a href="{{ site.baseurl }}/" alt="Hubble Enterprise">
 				<img src="{{ site.baseurl }}/assets/images/logo-title.svg" class="project-logo" width="3rem" height="3rem">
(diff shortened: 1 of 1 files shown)


parkr:
@larsxschneider I'm not sure what versions of GHE you wish to support & I'm not entirely sure what the desired outcome of this is, but we have shipped (for a while) [a number of `site.github.*` attributes](https://help.github.com/articles/repository-metadata-on-github-pages/) which might be useful here.

GitHugop:
Something like this can be used:

{{ site.url | replace: '://pages', '://' | remove_first: "https://" | remove_first: "http://" }}

larsxschneider:
@GitHugop Great suggestion, thank you! I'll use that 👍 

@parkr That looks very useful, especially for #13 . Thanks for the pointer!

PR #24, opened 2017-10-19
Readability of graph descriptions

brntbeer:
👋  Autodesk, @pluehne and @larsxschneider

I noticed when i was looking through the graphs ( 😍  they're amazing ) that the description of some of the graphs were underneath the graph themselves, and some graphs can be quite long (`users-git-versions` specifically comes to mind). My proposal is to move these short descriptions (where applicable) to the top of the graph so they don't need to scroll to the bottom of the graph.

## TODO
- [x] convert spaces back to tabs (sorry!)

### DIFF
--- a/docs/assets/css/stylesheet.css
+++ b/docs/assets/css/stylesheet.css
@@ -383,11 +383,13 @@ a
 .row
 {
 	width: 100%;
+	margin-bottom: 1rem;
 }
 
 .row .col-main
 {
 	width: 100%;
+	margin-bottom: 1rem;
 }
 
 .row .col-main svg
@@ -400,7 +402,14 @@ a
 .row .col-aside
 {
 	margin-top: 1rem;
+	margin-bottom: 1rem;
 	width: 100%;
+	clear: both;
+}
+
+.row .col-aside :last-child
+{
+	margin-bottom: 0;
 }
 
 .info-box
@@ -408,16 +417,6 @@ a
 	background-color: #f5f5f5;
 	border-radius: 0.5rem;
 	margin-bottom: 1rem;
-}
-
-.info-box:last-child
-{
-	margin-bottom: 0;
-}
-
-.row .col-aside:last-child
-{
-	margin-bottom: 0;
 }
 
 .info-box p
@@ -449,27 +448,43 @@ a
 
 	.row
 	{
-		display: flex;
-		align-items: flex-start;
+		clear: both;
 	}
 
-	.row-centered
+	.row::after
 	{
-		align-items: center;
+		content: "";
+		clear: both;
+		display: table;
 	}
 
 	.row .col-main
 	{
+		float: left;
+		clear: left;
 		width: 75%;
+		vertical-align: middle;
 		padding-right: 0.5rem;
 	}
 
 	.row .col-aside
 	{
+		vertical-align: top;
+		float: right;
+		clear: right;
 		width: 25%;
 		margin-top: 0;
 		padding-top: 0;
 		padding-left: 0.5rem;
+	}
+
+	.row-centered
+	{
+	}
+
+	.row-centered .col-aside
+	{
+		vertical-align: middle;
 	}
 }
 
--- a/docs/housekeeping-api-requests.html
+++ b/docs/housekeeping-api-requests.html
@@ -7,6 +7,13 @@ permalink: /housekeeping-api-requests
 <h3>API Requests</h3>
 
 <div class="row">
+  <div class="col-aside">
+		<div class="info-box">
+			<p>
+				The total number of <a href="https://developer.github.com/">GitHub API requests</a>.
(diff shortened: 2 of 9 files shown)


pluehne:
Thanks for reaching out to us 🙂. I agree that having the descriptions below some of the bigger charts is not optimal in terms of discoverability.

However, please note that the dashboard is designed in an (almost) responsive fashion. On larger screens, the descriptions are displays right of the charts.

I will look into whether some changes to the stylesheet would be required later when I have the time 🙂.

brntbeer:
> On larger screens, the descriptions are displays right of the charts.

@pluehne doh, i wondered if that was the case when it said `col-aside`. I should have looked deeper into the code. Let me know how I can help, I'd understand if you need to close this out! =)

pluehne:
@brntbeer: No worries 🙂. I think you have a point with the descriptions being hard to find on smaller screens! I will look how your changes look on different screen sizes and think about how to get handle the description texts on mobile devices.

Just know that our current support for mobile devices is not complete (most notably the collaboration graph), so there is some work to be done anyways 🙂.

brntbeer:
@pluehne my OCD may have gotten the best of me on this!

Not only did i put the tabs back, but i noticed if there was only `<div class="info-box">` under each page's `<h3>`, then the descriptions look nice regardless of the window size, while adhering to the original design of the information being on the side or under the graphs.

If that works for you, great. if not, and you want to keep this open until revisiting more changes with css, just let me know! no problem on my end 😄

pluehne:
@brntbeer: That sounds like a very good idea to me. Just give me some time to test your changes, as I'm currently not at my workstation 🙂.

And thanks again for caring about this issue!

pluehne:
@brntbeer: I found some time to test your changes 😃. However, the info boxes that you put on top now stretch across the entire width of the content:

screenshot from 2017-10-23 14-39-15

I think it would be best to have a responsive design where the first, introductory info box is displayed on the right side on big screens (like the other info boxes) and on top of the charts on small screens (unlike the others, which would stay below the chart).

What do you think?

pluehne:
@brntbeer: Thanks again for your pull request! After thinking some more about it, I would say that we should go for what I proposed yesterday—having an introductory info box above the chart on small screens and next to the chart with the other info boxes on large screens.

Would you like to work on this? If not or if you don’t have the time to, I could also finish this if you gave me permission to work on your pull request branch 😃. It’s really up to you!

I think that you have a good point with this pull request, which is why I’d like to merge your changes with the proposed revision at some point 😃.

brntbeer:
@pluehne was at a conference all day yesterday and didn't get a chance to have a think about this yet. Full disclosure: i think it'd be faster if you ended up working on it! I've checked the box to allow contributions from maintainers, so you should have access.

I agree with your proposal as well! I think that'd be great

pluehne:
@brntbeer: Thanks! I’ll make the changes then 😃.

pluehne:
Once more, CSS is a bit trickier than anticipated. I got the change mostly working in commit 4102aca by switching from flex back to oldschool float positioning (I currently adapted the API requests tab only). However, this makes it currently impossible to center the info boxes next to the collaboration chart vertically.

screenshot from 2017-10-24 19-07-33

I’ll spend some more time investigating this later …

pluehne:
@brntbeer: Just to keep you updated, I didn’t have the time yet to fix the remaining CSS issue with the collaboration chart.

If you or someone else knows a fix for that, I’d be happy to hear about it 😃.

brntbeer:
@pluehne ah, thanks for the update! I'll see if i can grab a friend to help as well =)

brntbeer:
@pluehne i've mentioned it to a few people. Do you mind if i push your changes to this branch?

pluehne:
@brntbeer: Absolutely not, go ahead 😃. I’d be curious to know how this issue can be solved cleanly!

brntbeer:
@pluehne felt like i had a solution in the vertical alignment before i remembered you mentioned the info box being the full length on `96rem` media was 👎 . 😄 

My solution involved setting `.row .col-aside` and  `.row .col-main` as `display: table-cell` and `.main-content` as `display: table-row`, but it left the top infobox either all the way on the left of the screen as it's own cell, or stretched across the top. whoops~

I'm not quite sure if there's a solution for this with smaller screens, and that's okay too!

larsxschneider:
What is the state here? What should we do?

I agree, that smaller screens are no priority. I expect most people to use Hubble on a desktop browser.

brntbeer:
Since this is dragging out a bit, my feelings won't be hurt if we close this down and re-think about it or wait to see if anyone else has this "problem".

Additionally, one small fix could just be to render a sort of warning on small screens that it's meant for wider displays

larsxschneider:
OK, let's close this for now!

PR #26, opened 2017-10-23
Improved user interaction with chart tooltips

pluehne:
The time series chart tooltips were a bit clunky.
The area in which the cursor had to be to invoke the tooltips was very small, and the tooltips
showed only one data entry at a time, making comparisons hard.

This commit changes that by showing tooltips already when the cursor hovers a date (hitting the exact data points is not necessary anymore), with all the data points available for that specific x position.

screenshot

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -80,6 +80,11 @@ var timeSeriesChartDefaults =
 			}
 		]
 	},
+	tooltips:
+	{
+		mode: "x",
+		intersect: false,
+	},
 };
 
 var barChartDefaults =


larsxschneider:
I would vote against this change as it makes it harder to inspect a single entry.

pluehne:
@larsxschneider: Granted. I’ll look whether there is a way to get both the combined tooltip view and the single one working depending on whether you hover a specific entry or just the date.

pluehne:
As I currently don’t have an idea how to implement this more nicely, I’ll close this pull request for now. I might come back to this later.

PR #28, opened 2017-10-25
Chart: repository activity

pluehne:
**This pull request is work in progress.**

This adds a report that records the number of active repositories in total, in organizations, and in user spaces for the last day, week, and four weeks, as well as corresponding charts in the *repositories* tab.

Again, I added some randomly generated demo data for [autodesk.github.io]().

This is how the new dashboard page looks like:

repository-activity-chart


pluehne:
Note that, in contrast to the previous charts, this chart records the last day, week, and *four weeks* (instead of 30 days). I decided to do so in order to have less fluctuations due to weekends. The past 30 days might include 4 *or* 5 weekends, whereas the past 4 weeks would always cover *exactly* 4 weekends.

We might consider doing the same with the other charts, however, this might affect already existing data and require a data reset for these charts 😐.

pluehne:
@toddocon: Moved discussion to here from #27.

What I wanted to say is that every day the data updater is invoked, the number of active repositories within the last day/week/4 weeks is recorded. Previously, we had aggregated that over the last 30 days.

There will still be a measurement for each day, and the data is cropped only after 2 years. I hope this clarifies a bit what I originally wanted to say 😃.

docs/repositories-activity.html
```diff
@@ -0,0 +1,45 @@
+---
```
larsxschneider: re: filename
we should agree on a scheme going forward (`repositories` vs `repos`)

toddocon: agreed - I'm working on the new organizations (vs orgs) one now. ;-)

pluehne: @toddocon: You might want to have a look at [https://github.com/Autodesk/hubble/pull/27#discussion_r147785133](our recent discussion on this topic), where we agreed to be consistent with the GitHub API naming scheme.

PR #31, opened 2017-10-29
Collaboration chart: draw chosen org first, make first org blue

larsxschneider:
If the collaboration chart of a single organization is visualized, then
ensure that the choosen org is always drawn first. This way the choosen
org is always drawn with the same color.

In addition, rotate the color wheel to start with the color "blue"
because I think it looks good.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -548,6 +548,28 @@ function visualizeSingleOrg(orgs, matrix, orgID) {
 		}
 	}
 
+	// Make the single org the first entry in the matrix to ensure that it
+	// always gets the same color in the chart.
+	if (orgID != 0) {
+		let t = orgs[orgID];
+		orgs[orgID] = orgs[0];
+		orgs[0] = t;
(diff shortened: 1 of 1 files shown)


docs/assets/js/charts.js
```diff
@@ -31,14 +34,11 @@ var extendedChartColors =
...
 	"#10d9be",
-	"#4c99e5",
-	"#c87adf",
-	"#ff5a99",
 ];
 
 function extendedChartColor(i)
 {
-	return extendedChartColors[(i + 3) % extendedChartColors.length];
+	return extendedChartColors[i % extendedChartColors.length];
```
larsxschneider: @pluehne I think this addition has no other purpose than rotating the color wheel, right?

pluehne: Exactly, and it shouldn’t be in the code.

pluehne:
Aside from this, I like the idea to make the selected organization blue in all cases very much. Maybe have a look at my other pull request, and after that let’s merge this one!

PR #36, opened 2017-11-08
List: organization owners

mlbright:
* ghe console script for all org owners
    * This was Patrick's reverse engineering efforts
    * We may have a ghe-console script from GitHub support
    (https://support.enterprise.github.com/hc/en-us/requests/51994)
    * We need to remove service accounts
* scaffolding for showing the above results

### DIFF
--- a/docs/org-owners.html
+++ b/docs/org-owners.html
@@ -4,7 +4,7 @@ title: Organization Owners
 permalink: /org-owners
 ---
 <h3>
-Organization owners
+Organization Owners
 </h3>
 
 <div class="row">
--- a/docs/org-owners.html
+++ b/docs/org-owners.html
@@ -1,0 +1,21 @@
+---
+layout: default
+title: Organization Owners
+permalink: /org-owners
+---
+<h3>
+Organization owners
+</h3>
+
+<div class="row">
+	<div class="col-main">
+		<div class="chart-container">
+		    <table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
+		</div>
+	</div>
+	<div class="col-aside">
+		<div class="info-box">
+			<p>Owners for each GitHub organization.</p>
+		</div>
+	</div>
+</div>
--- a/docs/org-owners.html
+++ b/docs/org-owners.html
@@ -8,14 +8,14 @@ Organization Owners
 </h3>
 
 <div class="row">
-    <div class="col-main">
-        <div class="chart-container">
-            <table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
-        </div>
-    </div>
-    <div class="col-aside">
-        <div class="info-box">
-            <p>Owners for each GitHub organization.</p>
-        </div>
-    </div>
+	<div class="col-main">
+		<div class="chart-container">
+			<table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
+		</div>
+	</div>
+	<div class="col-aside">
+		<div class="info-box">
+			<p>Owners for each GitHub organization.</p>
+		</div>
+	</div>
 </div>
--- a/docs/org-owners.html
+++ b/docs/org-owners.html
@@ -8,14 +8,14 @@ Organization Owners
 </h3>
 
 <div class="row">
-	<div class="col-main">
-		<div class="chart-container">
-		    <table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
-		</div>
-	</div>
-	<div class="col-aside">
-		<div class="info-box">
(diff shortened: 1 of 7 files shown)


toddocon:
Does this only run on the appliance? I've re-created your files in my environment (to test) and it's not working. I pull all of my stats via remoteRun so I think in org-owners.sh this line presumes running on the appliance, yes?

`github-env bin/runner -e production`

toddocon:
The error was:

`...
    self.executeScript(self.scriptPath("org-owners.sh"))
AttributeError: 'ReportOrgOwners' object has no attribute 'scriptPath'
`

So, in ReportOrgOwners.py, I changed this

`self.executeScript(self.scriptPath("org-owners.sh"))`

to this

`self.executeScript(os.path.join("scripts",` "org-owners.sh"))`

Now the data file is built. My output, though, is a table 10,221 rows long. There is one owner per line.

mlbright:
@toddocon I fixed the ghe-console script (thank you GHE support!)

mlbright:
I'm not sure why you need to specify 'scripts' and I don't.

docs/org-owners.html
```diff
@@ -0,0 +1,21 @@
+---
+layout: default
+title: Organization Owners
+permalink: /org-owners
+---
+<h3>
+Organization owners
```
pluehne: @mlbright: Can you make this title case (“Organization Owners”)?

updater/update-stats.py
```diff
@@ -21,6 +21,7 @@
 from reports.ReportRepositoryHistory import *
 from reports.ReportTokenlessAuth import *
 from reports.ReportUsers import *
+from reports.ReportOrgOwners import *
```
pluehne: @mlbright: Could you please sort these alphabetically?

updater/update-stats.py
```diff
@@ -82,6 +83,7 @@ def main():
 	ReportRepositoryHistory(configuration, dataDirectory, metaStats).update()
 	ReportTokenlessAuth(configuration, dataDirectory, metaStats).update()
 	ReportUsers(configuration, dataDirectory, metaStats).update()
+	ReportOrgOwners(configuration, dataDirectory, metaStats).update()
```
pluehne: @mlbright: Could you please sort these alphabetically?

pluehne:
@toddocon: It’s weird that

```python
self.executeScript(self.scriptPath("org-owners.sh"))
```

didn’t work for you, but

```python
self.executeScript(os.path.join("scripts", "org-owners.sh"))
```

did. In fact, `self.scriptPath` should eventually do the same thing.

Also, other reports (such as for Git versions) use the same call. Are they broken for you as well?

pluehne:
For future versions of this report, I’d like the following improvements:

- a search box to quickly find the owners of a specific organization (which, I think, is the typical use case of this report)
- links to the user accounts of the owners, just like the organizations themselves are linked

docs/org-owners.html
```diff
@@ -0,0 +1,21 @@
...
+permalink: /org-owners
+---
+<h3>
+Organization Owners
+</h3>
+
+<div class="row">
+	<div class="col-main">
+		<div class="chart-container">
+		    <table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
```
pluehne: @mlbright: You accidentally mixed spaces and tabs here.

docs/org-owners.html
```diff
@@ -8,14 +8,14 @@
...
-		</div>
-	</div>
-	<div class="col-aside">
-		<div class="info-box">
-			<p>Owners for each GitHub organization.</p>
-		</div>
-	</div>
+    <div class="col-main">
+        <div class="chart-container">
+            <table class="table" data-url="{{ site.dataURL }}/org-owners.tsv"></table>
```
pluehne: Sorry, I should have been more specific. We’re using tabs in the other files, and it would be best to keep that consistent.

pluehne:
As I don’t have write access to @mlbright’s branch, I [branched this pull request](https://github.com/Autodesk/hubble/tree/patrick/org-owners).

7e9ce8dac3ff762b47181994d3711950d02fc0f2 adds some demo data for [autodesk.github.io](), and be7a159581fc6f833d9e63239651751adc396f32 is a work-in-progress commit, where I started to support multiple owner names on a single line in the table (with links to their GitHub profiles).

Note that the updater report’s output format needs to be changed in the same way (instead of having one line for each (organization, owner) pair, just output a single line for each organization with all the owners comma-separated). Also, the column heading should be *owners* and not *owner* (singular).

larsxschneider:
Superseded by #38

PR #39, opened 2017-11-19
Chart: total number of organizations and teams

toddocon:
This is a new history chart under ORGANIZATIONS


updater/reports/ReportOrgHistory.py
```diff
@@ -0,0 +1,55 @@
...
+# Lists the total number of organizations
+class ReportOrgHistory(ReportDaily):
+	def name(self):
+		return "org-history"
+
+	def updateDailyData(self):
+		newHeader, newData = self.parseData(
+			self.executeQuery(self.query())
+		)
+		self.header = newHeader if newHeader else self.header
```
larsxschneider: `self.header` is not previously defined. Therefore, we would not set anything in the else part, right?

updater/reports/ReportOrgHistory.py
```diff
@@ -0,0 +1,55 @@
...
+	def subquery(self, userType):
+		query = '''
+			SELECT
+				COUNT(*) AS count
+			FROM
+				users
+			WHERE
+				users.type = "Organization"
+			AND
+				users.login <> "github-enterprise"
```
larsxschneider: Good catch! We need to ignore this invalid organization in other reports, too. I think it would be easiest to hard code "github-enterprise" to https://github.com/Autodesk/hubble/blob/master/updater/reports/Report.py#L198-L210

//cc @pluehne ?

updater/reports/ReportOrgHistory.py
```diff
@@ -0,0 +1,55 @@
...
+		return query
+
+	# Collects the number of teams
+	def totalTeamsQuery(self, userType):
+		query = '''
+			SELECT
+				COUNT(*) AS count
+			FROM
+				teams
+			'''
```
larsxschneider: Isn't that the total number of teams? Don't we want to know the number of teams per org here?
I dunno if we should mix that into this repo. @pluehne ?

larsxschneider: As I hit "send", I realized you really want to count the total number of teams here 😄

larsxschneider:
Closed and continued in #43

PR #40, opened 2017-11-20
fix executeGHEConsole() / ReportOrgOwners for Hubble service

larsxschneider:
The executeGHEConsole() function (and consequently the ReportOrgOwners)
was broken if Hubble is configures as service that runs on the GitHub
Enterprise appliance. Fix it.


updater/reports/Report.py
```diff
@@ -80,19 +80,28 @@ def scriptPath(self, script):
...
-				with open(script) as f:
-					assert(stdin == None)
-					stdin = f.read()
-				command = ["bash -s", "--"]
-			except:
-				command = script
+			if not stdin:
+				# A script is either a path to a file or a list of strings
+				# with bash commands. Both are executed by passing the script
+				# to stdin to `bash -s --`. This method only works if we habve
```
pluehne: typo: “habve”

updater/reports/Report.py
```diff
@@ -80,19 +80,28 @@ def scriptPath(self, script):
...
 		if self.configuration["remoteRun"]["enabled"]:
-			try:
-				with open(script) as f:
-					assert(stdin == None)
-					stdin = f.read()
-				command = ["bash -s", "--"]
-			except:
-				command = script
+			if not stdin:
+				# A script is either a path to a file or a list of strings
```
pluehne: Maybe we should put this explanation in the comment above `executeScript` so that this is immediately clear when reading the function? And I wouldn’t mind having this information here as well.

PR #43, opened 2017-11-20
Chart: total number of organizations and teams (v2)

larsxschneider:
This is a continuation of @toddocon work in #39


updater/config.py.example
```diff
@@ -54,10 +54,15 @@ configuration = \
...
 	# Certain entities (organizations or repositories) might exist only for
-	# backup purposes. If you have a special naming scheme for them, then
-	# define that here to exclude them from all kinds of statistics.
-	# "excludedEntities": "discarded-%",
-
+	# backup or testing purposes. If you have a special naming scheme for them,
+	# then define that here to exclude them from all kinds of statistics.
+	# "github-enterprise" is a default organization in GitHub Enterprise that
+	# we exclude here, too (Note: this also excludes repositories with the name
+	" "github-enterprise").
```
pluehne: The leading `"` looks like a typo, or is this intentional?

updater/reports/ReportOrgHistory.py
```diff
@@ -0,0 +1,27 @@
+from .ReportDaily import *
+
+# Lists the total number of organizations
+class ReportOrgHistory(ReportDaily):
+	def name(self):
+		return "orgs-history"
```
pluehne: If the report file is called `orgs-history`, shouldn’t we rename the class to `ReportOrgsHistory` then?

pluehne: However, now I think it might be best to call the output file and the report `org-history` and `ReportOrgHistory` (because it really sounds strange in plural here), and only prepend `orgs-` to the HTML pages, which are in the category `orgs`. What do you think, @larsxschneider?

updater/reports/ReportTeamHistory.py
```diff
@@ -0,0 +1,27 @@
+from .ReportDaily import *
+
+# Lists the total number of organizations
+class ReportTeamHistory(ReportDaily):
+	def name(self):
+		return "teams-history"
```
pluehne: See above comments for `org-history`.

toddocon:
This chart mimics the output from the stats API call ( api/v3/enterprise/stats/orgs ). It shows the total number of orgs and total number of teams. The API also returns total number of disabled Orgs and total team members. The first is always 0 for us so I skipped it. The second, total team members, I have not been able to reverse engineer via a SQL call. 

If we want total teams per Org, then maybe that should be a tabled below this chart? As a v2 enhancement to this chart.

toddocon:
Yes, in the current Organization Owners list, the built-in github-enterprise Org/User shows up. To exclude this as Lars has done is correct. I missed seeing this in my Org list since there are so many but now that I've checked, it's there. 

ghe

larsxschneider:
OK. Not sure if total members of teams per org is necessary. That was just a misunderstanding on my end. I divided the two numbers into two graphs because they can vastly differ (e.g. you could have only 5 orgs but 5000 teams) and then they would be hard to see.

larsxschneider:
@toddocon You need to adjust your local config file to exclude "github-enterprise".

larsxschneider:
@toddocon Can you give me a quick review for the final PR here?

toddocon:
I will adjust my file.

My org/team numbers do differ drastically. Good catch on breaking the data up into two charts. Thanks!

toddocon:
Reviewing now....

PR #44, opened 2017-11-20
fix Git download chart for GHE 2.11.0

larsxschneider:
The "github-audit.log" log file introduced in GHE 2.11.0 is only rolled
once a week. This was reported as bug and is likely fixed in an upcoming
version. In the meantime we grep for all log entries that have been
written yesterday.

### DIFF
--- a/updater/scripts/git-download.sh
+++ b/updater/scripts/git-download.sh
@@ -15,13 +15,20 @@ function ghe_greater_equal () {
 
 if ghe_greater_equal "2.11.0" ; then
     LOG_FILE="/var/log/github-audit.log"
+    # The "github-audit.log" log file introduced in GHE 2.11.0 is only rolled
+    # once a week. This was reported as a bug and is likely fixed in an
+    # upcoming version. In the meantime we grep for all log entries that have
+    # been written yesterday.
+    GREP_YESTERDAY="grep -F '$(date --date='yesterday' +'%b %d')'"
 else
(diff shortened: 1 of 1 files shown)


updater/scripts/git-download.sh
```diff
@@ -15,13 +15,20 @@ function ghe_greater_equal () {
 
 if ghe_greater_equal "2.11.0" ; then
     LOG_FILE="/var/log/github-audit.log"
+    # The "github-audit.log" log file introduced in GHE 2.11.0 is only rolled
+    # once a week. This was reported as bug and is likely fixed in an upcoming
```
pluehne: Typo: “as a bug”

updater/scripts/git-download.sh
```diff
@@ -15,13 +15,20 @@ function ghe_greater_equal () {
...
+    # version. In the meantime we grep for all log entries that have been
+    # written yesterday.
+    GREP_YESTERDAY="grep -F '$(date --date='yesterday' +'%b %d')'"
```
pluehne: Maybe, we could add something like `# TODO: remove workaround once GitHub Enterprise is patched` so that we clean this up as soo as this workaround isn’t necessary anymore?

larsxschneider: Well, I don't think we can remove it because GHE 2.11.0 - 2.11.x will always have this bug.

PR #46, opened 2017-11-21
Fix Git download script

pluehne:
This fixes an issue where the log files read by Hubble were empty, which is why the Git download report got 0 traffic as a result.

This lets the Git download script check not only for log entries in the previous Git audit log but also the current one. Because of the unwanted log rotation change introduced in GitHub Enterprise 2.11, we can’t assume that the log files searched for with .1* contain yesterday’s information (it might still reside in the current, not-yet-rotated log).

I tested this on a GitHub Enterprise 2.11.2 machine, and I’m confident that this doesn’t break 2.10 support.

### DIFF
--- a/updater/scripts/git-download.sh
+++ b/updater/scripts/git-download.sh
@@ -14,23 +14,20 @@ function ghe_greater_equal () {
 }
 
 if ghe_greater_equal "2.11.0" ; then
-    # check the two most recent log files
-    CAT_LOG_FILE="zcat -f /var/log/github-audit.log{,.1*}"
     # The "github-audit.log" log file introduced in GHE 2.11.0 is only rolled
     # once a week. This was reported as a bug and is likely fixed in an
-    # upcoming version. In the meantime we grep for all log entries that have
-    # been written yesterday.
-    GREP_YESTERDAY="grep -F '$(date --date='yesterday' +'%b %d')'"
+    # upcoming version. In the meantime, we grep for all log entries in the two
+    # most recent log files (because the information from yesterday may or not
+    # be rotated already).
+    CAT_LOG_FILE="zcat -f /var/log/github-audit.log{,.1*} | grep -F '$(date --date='yesterday' +'%b %d')'"
 else
     # check yesterday’s log file
     CAT_LOG_FILE="zcat -f /var/log/github/audit.log.1*"
-    GREP_YESTERDAY="tee"
 fi
 
 echo -e "repository\tuser\tcloning?\trequests/day\tdownload/day [B]"
 
 eval "$CAT_LOG_FILE" |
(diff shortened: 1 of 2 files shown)


updater/scripts/git-download.sh
```diff
@@ -14,20 +14,22 @@ function ghe_greater_equal () {
...
-    LOG_FILE="/var/log/github/audit.log"
+    # check yesterday’s log file
+    CAT_LOG_FILE="zcat -f /var/log/github/audit.log.1*"
     GREP_YESTERDAY="tee"
 fi
 
 echo -e "repository\tuser\tcloning?\trequests/day\tdownload/day [B]"
 
-zcat -f $LOG_FILE.1* |
+eval "$CAT_LOG_FILE" |
```
larsxschneider: Why do we need eval here? `zcat -f $LOG_FILE` wouldn't work?

pluehne: Yes, it wouldn’t, because the Bash wouldn’t expand the wildcards within `CAT_LOG_FILE` without `eval`.

larsxschneider: I guessed that. But in that case we can move the $GREP_YESTERDAY into this eval, no?

pluehne: Yep, we should be able to put both `$CAT_LOG_FILE` and `$GREP_YESTERDAY` into the same `eval`. I just thought it might be nicer to keep the commands individual and not have pipe operators within `eval`s and outside of them.

What do you think?

larsxschneider: All this stuff is just a workaround and the workaround requires both evals. Therefore, I think it should be one eval in the end. As soon as the workaround is not necessary anymore we can get rid of the eval completely.

pluehne: All right, I’ll commit the change 🙂.

updater/scripts/git-download.sh
```diff
@@ -14,23 +14,20 @@ function ghe_greater_equal () {
...
-    # been written yesterday.
-    GREP_YESTERDAY="grep -F '$(date --date='yesterday' +'%b %d')'"
+    # upcoming version. In the meantime, we grep for all log entries in the two
+    # most recent log files (because the information from yesterday may or not
+    # be rotated already).
+    CAT_LOG_FILE="zcat -f /var/log/github-audit.log{,.1*} | grep -F '$(date --date='yesterday' +'%b %d')'"
 else
     # check yesterday’s log file
     CAT_LOG_FILE="zcat -f /var/log/github/audit.log.1*"
-    GREP_YESTERDAY="tee"
```
pluehne: Note that we can now get rid off this `tee` call when merging both `eval`s into one 😃.

PR #57, opened 2017-12-06
Improve PR usage explanation

larsxschneider:


docs/pr-usage.html
```diff
@@ -22,7 +22,8 @@ <h3>Pull Request Usage</h3>
 				The percentage of active repositories located in organizations, having more than one contributor, and using pull requests for contributions.
 			</p>
 			<p>
-				In this context, <em>active</em> designates repositories with at least two pushes in the last four weeks.
+				In this context, <em>active</em> designates repositories with at least two users that have pushed commits in the last four weeks.
+				The rational behind this is that pull request usage is most important in a collaborative environment.
```
pluehne: It should be “rationale” (noun) and not “rational” (adjective).

larsxschneider: Thanks for explaining the difference. I wasn't aware of it before 👍

pluehne: You’re welcome!

Issue #58, opened 2017-12-12
Feature request: Test suite/linter


filmaj:
- `docs/assets/charts.js` is pretty big and as I look to contribute to the project, I would like to ensure I'm not breaking anything in there.
- adding a set of explicit JS linter rules for `charts.js` would be nice. I mentioned this for selfish reasons as my IDE currently looks like a red underline mess of spaghetti due to this project's preference for tabs 😜

pluehne:
I totally agree we need a consistent coding style. @larsxschneider and I will discuss this tomorrow!

Aside from that, we’re definitely interested in having automated tests (especially for the updater, we didn’t think about testing the JavaScript side much yet). If you have suggestions about how best to test JavaScript functionality, we’d like to know about that 😃.

filmaj:
>  If you have suggestions about how best to test JavaScript functionality, we’d like to know about that 😃.

Cool! I do have some suggestions on testing JavaScript, I love this topic 💃. I think all of it depends on how we want to be covered, how much code coverage we want, and how much tooling we're willing to have contributors require.

# How Much Tooling

We could leverage (and, implied here is: it will be required for developers / potential contributors) [node.js](http://nodejs.org), which gives us a JS runtime from the command-line. From there, we could formalize any config, dependencies and common project tasks (such as linting and testing) related to the web app / JS in a `package.json` file (a.k.a. node's `requirements.txt`). It'd be easy to formalize which test runner and linter we want to use - I'd suggest [jasmine](https://jasmine.github.io/) and [eslint](https://eslint.org) for those just because I have the most familiarity with them, but there are many to choose from and as long as this project picks one, that's what I care about most :). We can also encapsulate the test running and linting commands by defining "scripts" inside `package.json`. I recently worked on adding something similar to a JS project at Adobe - you can check out [the `package.json` for that](https://github.com/adobe/jsonschema2md/blob/master/package.json) for a sense of what I'm talking about. In particular the [`devDependencies` section](https://github.com/adobe/jsonschema2md/blob/master/package.json#L28) (which defines what dev tools we need like test runners and linters) as well as the [`scripts` section](https://github.com/adobe/jsonschema2md/blob/master/package.json#L9) (which defines what [commands we can run](https://github.com/adobe/jsonschema2md#tests)). With this setup, a potential contributor could clone the repo, and as long as they have node.js installed, could theoretically:

    $ cd docs
    $ npm install # use `npm` which comes with node.js to install any dependencies
[...]

PR #59, opened 2017-12-13
Fixes #20: adds a spinner during graph load

### LINKED ISSUE #20: Display "Loading" label in collaboration chart
The "Collaboration Across Organizations" chart takes a bit of time on the first load if many organizations are visualized. A "Loading..." label would be nice to inform the user that things are happening.

filmaj:
This PR addresses #20: display a loading indicator while Hubble is off fetching data.

# Changes

- Leverages [spin.js](http://spin.js.org/) for the spinner, so I added it to the layout template
- The main column element's CSS had to be tweaked to use relative positioning
- Each function creating each type of graph was tweaked to:
  a) create a spinner
  b) hide the spinner once network request completes

# Considerations

- What to do if the async net request for data errors? Currently, nothing happens, so the spinners keep on spinning.

# Testing It

I pointed the `_config.yml` file to `http://localhost:4000/demo-data`, then ran `jekyll serve` from the `docs/` folder. I loaded up http://localhost:4000 in Chrome, and using the Chrome Developer Tools' Network tab, I set the simulated network speed to "Slow 3G", then refreshed the page. That network profile was slow enough for me to be able to ensure that the spinner does show up relatively centered over where incoming graphs will be rendered to.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -115,10 +115,18 @@ var stackedBarChartDefaults =
 	maintainAspectRatio: false
 };
 
+function createSpinner(canvas)
+{
+	var parent = $("<div style=\"position:absolute;height:100%;width:100%;\"></div>");
+	parent.insertBefore($(canvas));
+	return new Spinner().spin(parent[0]);
+}
+
 function createHistoryChart(canvas)
 {
 	var url = $(canvas).data("url");
 
+	var spinner = createSpinner(canvas);
 	d3.tsv(url,
 		function(row)
 		{
@@ -228,6 +236,8 @@ function createHistoryChart(canvas)
 					},
 					options: timeSeriesChartDefaults
 				});
+		}).on("load.spinner", function() {
+			spinner.stop();
 		});
 }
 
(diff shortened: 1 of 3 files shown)


docs/assets/js/charts.js
```diff
@@ -115,10 +115,18 @@ var stackedBarChartDefaults =
 	maintainAspectRatio: false
 };
 
+function createSpinner(canvas)
+{
+	var parent = $("<div style=\"position:absolute;height:100%;width:100%;\"></div>");
```
larsxschneider: Here and in the other places below I think we should rather use `let` to declare the variables, no? I know this is not yet consistent throughout the codebase (... and that's why I love you [linting proposal](#58)).

Does this sound good to you? If yes, then you don't need to to anything. I would just adjust your commit on merge.

filmaj: Sounds fine to me! Whatever style the project decides on, I will conform to. 👍 

Once I start putting together the JS test running and linting changes I have planned, we can discuss in finer detail the particular JS style rules the project wants to adopt.

larsxschneider:
Merged via ada9a070acece2327bc4bbb45d1d056cca158089
Thank you 👍

PR #63, opened 2017-12-14
Fixes #62: spinner container gets in the way of mouseover events with charts

### LINKED ISSUE #62: No more mouse over?
I just updated to the latest Hubble and now I no longer have mouse over information like mousing over bubbles or the bands in the collaboration charts.

I also reloaded the site in Safari in a new private window, I see the same behavior.

filmaj:
Sorry about the bug with the spinner, my bad!

I've fixed that in this pull request by ensuring we remove the container housing the spinner from the DOM after calling `stop` on it. I also added a test for this to our new test framework to ensure we don't regress on this in the future :)


larsxschneider:
Thank you 👍

Issue #10, opened 2017-10-13
Move favicons to `docs/assets/images`?


larsxschneider:
Can/Should we move `docs/favicon*.png` and `docs/apple-touch-icon*.png` to `docs/assets/images`?

pluehne:
I’m not sure about this one, because I think that browsers will look for the default filenames in the root directory.

On the other side, I linked the favicons in the HTML `<head>` … I’ll investigate whether moving the favicons to `docs/assets/images` would cause trouble.

filmaj:
We can certainly use the `href` attribute for the `<link>` elements to specify the icons in subdirectories. For the apple icons, according to [Apple's documentation on this](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4), it's true that the root directory will be searched for appropriately-named files, but only if the page does not specify `<link>` elements describing where to look for the icons. In our case, since we have the `<link>`s, that wouldn't apply to us.

pluehne:
Thanks for investigating the favicons issue. I also assumed that `<link>`s would be sufficient to let all kinds of browsers know where to look for favicons/touch icons, but I didn’t find a reliable source and didn’t have the time to search further. I also like the idea of cleaning up the root directory, so let’s move this as you proposed in #68!

pluehne:
Fixed by #68 and f16f3de50ce8a67cd6992ac2babe490dee3b474b.

PR #68, opened 2017-12-19
Move favicons to assets/images

filmaj:
As per #10.

I like this change as it clears up the root webapp directory a little bit.


pluehne:
Thanks for the pull request! I like this very much, and I’d even suggest putting the icon files in `assets/icons` to separate them from logos, background images, etc. Would that make sense to you?

pluehne:
@larsxschneider and I decided to move the icons into `assets/icons`.

I made the necessary change and merged your pull request as per f16f3de50ce8a67cd6992ac2babe490dee3b474b 😃.

Issue #71, opened 2017-12-21
Updater service not triggered without user session


mzzmjd:
When using the hubble-enterprise.timer on the appliance server, statistics are never collated. It seems that when you log off the primary server the updater stops too.

```[ghebackup@ghebackup ~]$ ssh -p 122 admin@github
     ___ _ _   _  _      _      ___     _                    _
    / __(_) |_| || |_  _| |__  | __|_ _| |_ ___ _ _ _ __ _ _(_)___ ___
   | (_ | |  _| __ | || | '_ \ | _|| ' \  _/ -_) '_| '_ \ '_| (_-</ -_)
    \___|_|\__|_||_|\_,_|_.__/ |___|_||_\__\___|_| | .__/_| |_/__/\___|
                                                   |_|

Administrative shell access is permitted for troubleshooting and performing
documented operations procedures only. Modifying system and application files,
running programs, or installing unsupported software packages may void your
support contract. Please contact GitHub Enterprise technical support at
enterprise@github.com if you have a question about the activities allowed by
your support contract.
Last login: Thu Dec 21 07:11:59 2017 from [masked-ip-address]
admin@github-primary:~$ journalctl -f --user-unit hubble-enterprise.timer
-- Logs begin at Wed 2017-06-21 20:01:19 UTC. --
Dec 21 06:59:43 github-primary systemd[23519]: Starting Runs Hubble Enterprise updater periodically.
Dec 21 06:59:43 github-primary systemd[23519]: Started Runs Hubble Enterprise updater periodically.
Dec 21 07:11:56 github-primary systemd[23519]: Stopping Runs Hubble Enterprise updater periodically.
Dec 21 07:11:56 github-primary systemd[23519]: Stopped Runs Hubble Enterprise updater periodically.
Dec 21 07:11:59 github-primary systemd[16682]: Starting Runs Hubble Enterprise updater periodically.
Dec 21 07:11:59 github-primary systemd[16682]: Started Runs Hubble Enterprise updater periodically.
Dec 21 07:13:16 github-primary systemd[16682]: Stopping Runs Hubble Enterprise updater periodically.
Dec 21 07:13:16 github-primary systemd[16682]: Stopped Runs Hubble Enterprise updater periodically.
Dec 21 07:16:13 github-primary systemd[22010]: Starting Runs Hubble Enterprise updater periodically.
Dec 21 07:16:13 github-primary systemd[22010]: Started Runs Hubble Enterprise updater periodically.
```
[...]

PR #72, opened 2018-01-03
Switch updater to systemd system service

pluehne:
This makes the `systemd` system service a user service. There are two reasons that motivated this change.

1. User services require an active session of the specified user. This has the drawback that the updater would not be triggered after reboots if no SSH connection to the GitHub Enterprise appliance was made (see also #71, which should be fixed by this pull request).
1. User services cannot depend on system services. Because of that, the updater service could not reliably wait for the rest of GitHub Enterprise to start.

### DIFF
--- a/updater/README.md
+++ b/updater/README.md
@@ -34,12 +34,12 @@ The updater may be run in two modes:
 1. Create a config file in `/opt/autodesk/hubble-enterprise/config.py` (based on `[config.py.example](config.py.example)`).
 1. Enable and start the **periodic** updater process (persistently, survives reboots):
 	```sh
-	systemctl --user enable hubble-enterprise.timer
-	systemctl --user start hubble-enterprise.timer
+	sudo systemctl enable hubble-enterprise.timer
+	sudo systemctl start hubble-enterprise.timer
 	```
 1. Optionally, you may **trigger** the updater process (just once):
 	```sh
-	systemctl --user start hubble-enterprise.service
+	sudo systemctl start hubble-enterprise.service
 	```
 
 #### Upgrading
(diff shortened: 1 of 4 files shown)


updater/README.md
```diff
@@ -72,7 +72,7 @@ The updater may be run in two modes:
 
 The updater’s log is accessible as follows:
 ```sh
-journalctl -f --user-unit hubble-enterprise.service
+journalctl -fu hubble-enterprise.service
```
larsxschneider: this change is unrelated, right?

updater/packaging/debian/templates/Makefile
```diff
@@ -2,7 +2,7 @@ all: # nothing to build
 
 PKGNAME = hubble-enterprise
 INSTALLDIR = $(DESTDIR)/opt/autodesk/$(PKGNAME)
-SYSTEMDDIR = $(DESTDIR)/home/admin/.local/share/systemd/user
+SYSTEMDDIR = $(DESTDIR)/usr/lib/systemd/system
```
larsxschneider: Do you see a way to move our stuff into a subdirectory here? Would that make sense?

updater/README.md
pluehne: No, it isn’t. `-u` stands for `--unit` and refers to system units, while `--user-unit` is the equivalent for user units and doesn’t have a corresponding shorthand option.

updater/packaging/debian/templates/Makefile
pluehne: I think that `systemd` doesn’t look for unit files in a subdirectory.

PR #73, opened 2018-01-08
Fix dh_make call for Debian packaging

pluehne:
`dh_make` recently changed from Python 2 to Python 3. For this reason, the call to `dh_make` with Python 2 won’t work on systems with new `dh_make` versions. Instead of locking the Python version to Python 2, call Python without the fixed version number. This should be safe, because we can expect Linux distributions to ship a version of `dh_make` that works with the system’s default Python version.

### DIFF
--- a/updater/packaging/debian/Makefile
+++ b/updater/packaging/debian/Makefile
@@ -12,7 +12,7 @@ all:	clean
 	cd $(BUILDDIR) && \
 	export DEBEMAIL=$(EMAIL) && \
 	export DEBFULLNAME=$(MAINTAINER) && \
-	python2 /usr/bin/dh_make -p $(PKGNAME)_$(PKGVERSION) -i --native -c mit -y && \
+	python /usr/bin/dh_make -p $(PKGNAME)_$(PKGVERSION) -i --native -c mit -y && \
 	dpkg-buildpackage -rfakeroot -d -us -uc
 	mkdir -p $(PKGDIR)
 	mv $(BUILDDIR)/../*.deb $(PKGDIR)/


updater/packaging/debian/Makefile
```diff
@@ -12,7 +12,7 @@ all:	clean
 	cd $(BUILDDIR) && \
 	export DEBEMAIL=$(EMAIL) && \
 	export DEBFULLNAME=$(MAINTAINER) && \
-	python2 /usr/bin/dh_make -p $(PKGNAME)_$(PKGVERSION) -i --native -c mit -y && \
+	python /usr/bin/dh_make -p $(PKGNAME)_$(PKGVERSION) -i --native -c mit -y && \
```
larsxschneider: What would happen if I have the wrong Python version? Would it scream loud or error silently?

pluehne: Calling `dh_make` with the wrong Python version results in errors.

`dh_make` 2.201701 (the new version) with Python 2 gives a module dependency error:

```
Traceback (most recent call last):
  File "/usr/bin/dh_make", line 16, in <module>
    from enum import Enum
ImportError: No module named enum
```

`dh_make` 2.201608 (the old version) with Python 3 even results in a syntax error:

```
  File "dh_make", line 715
    os.mkdir('source', 0755)
                          ^
SyntaxError: invalid token
```

In both cases, we get an error.

pluehne: However, we should, perhaps, require `dh_make` to succeed for any of the later commands.

pluehne: By the way, I’d like to call `dh_make` without `python` and to rely on a proper shebang declaration. However, `dh_make` as of 2.201608 declares `/usr/bin/python` without specifying the required Python version explicitly. This is unfortunate, because now, we have to care about the correct version. However, `dh_make` 2.201701 has a proper interpreter directive (`/usr/bin/python3`).

PR #77, opened 2018-01-10
Provide generalized aggregation

pluehne:
This implements a generalized method for aggregating time-series data. Data can be aggregated over week or month intervals with a variety of aggregation methods to choose from.

This will be useful for providing chart views at different levels (such as two-year periods vs. just showing the last month). Additionally, the generalized form of aggregation can be used to smooth out graphs where the sampling frequency changed with an update to Hubble Enterprise.

The aggregation is done by splitting the time data into subsequent, gapless periods of time (weeks starting with Mondays or months), for each of which the aggregated values are then computed and returned.

Aggregation methods define how to aggregate the values within individual time periods. The following aggregation methods are supported:

- sum
- mean
- min
- max
- first (the chronologically first available value for that period)
- last
- median

Periods with incomplete data at the beginning or the end of the time series are excluded from the aggregation.

Finally, the pull request usage chart is changed to make use of the new aggregation facilities to reduce the granularity from daily to monthly data for now. This might be changed when we implement detail views.

I also added several unit tests to check the aggregation methods (for off-by-one errors in particular) as well as a short piece of documentation on the new configuration options.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -142,6 +142,111 @@ function createSpinner(canvas)
     };
 }
 
+function aggregateTimeData(data, aggregationConfig)
+{
+    if (!(data instanceof Array))
+        throw 'expected data array as input';
+
+    if (data.length < 1)
+        return;
+
+    // Turn date strings into proper date objects
+    for (let i = 0; i < data.length; i++)
+        data[i]['date'] = d3.isoParse(data[i]['date']);
+
+    // Sort data, just in case it isn’t already
+    data.sort((row1, row2) => row1['date'] - row2['date']);
+
+    const dateStart = data[0]['date'];
+    // Ranges are exclusive, so add one more day to include the last date
+    const dateEnd = d3.utcDay.offset(data[data.length - 1]['date'], 1);
+
+    let period;
+
+    switch (aggregationConfig['period'])
+    {
+        case 'week':
+            period = d3.utcMonday;
+            break;
+        case 'month':
+            period = d3.utcMonth;
+            break;
+        default:
+            throw 'unknown aggregation period "' + aggregationConfig['period'] + '"';
+    }
+
+    // Don't use incomplete periods at the beginning and the end of the data
+    const t0 = period.ceil(dateStart);
+    // In d3, ranges include the start value but exclude the end value.
+    // We want to include the last period as well, so add one more period
+    const t1 = period.offset(period.floor(dateEnd), 1);
+    const periods = period.range(t0, t1);
+
+    let aggregatedData = Array();
+
+    for (let i = 0; i < periods.length - 1; i++)
+    {
+        const t0 = periods[i];
+        const t1 = periods[i + 1];
+
+        // Note that this assumes complete data in the period.
+        // Should data points be missing, aggregation methods such as the sum will lead to results that can't be
+        // compared to periods with complete data.
+        // Hence, the maintainers of the data need to ensure that the input is well-formed
+        const dates = data.filter(row => row['date'] >= t0 && row['date'] < t1);
+
+        let row = Object();
+        row['date'] = t0;
+
+        $.each(Object.keys(data[0]),
+            function(keyID, key)
+            {
+                // Exclude the date itself from aggregation
+                if (key == 'date')
+                    return;
+
+                if (dates.length == 0)
+                {
+                    row[key] = undefined;
+                    return;
+                }
+
+                const accessor = (row => row[key]);
+
+                switch (aggregationConfig['method'])
+                {
+                    case 'sum':
+                        row[key] = d3.sum(dates, accessor);
+                        break;
+                    case 'mean':
+                        row[key] = d3.mean(dates, accessor);
+                        break;
+                    case 'median':
+                        row[key] = d3.median(dates, accessor);
+                        break;
+                    case 'first':
+                        row[key] = dates[0][key];
+                        break;
+                    case 'last':
+                        row[key] = dates[dates.length - 1][key];
+                        break;
+                    case 'min':
+                        row[key] = d3.min(dates, accessor);
+                        break;
+                    case 'max':
+                        row[key] = d3.max(dates, accessor);
+                        break;
+                    default:
+                        throw 'unknown aggregation method "' + aggregationConfig['method'] + '"';
+                }
+            });
+
+        aggregatedData.push(row);
+    }
+
+    return aggregatedData;
+}
+
 function createHistoryChart(canvas)
 {
     const url = $(canvas).data('url');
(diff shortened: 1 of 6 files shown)


docs/_layouts/default.html
```diff
@@ -27,7 +27,7 @@
 		<script src="{{ site.baseurl }}/assets/js/vendor/moment-with-locales.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/Chart-2.7.1.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/spin-2.3.2.min.js"></script>
-		<script src="{{ site.baseurl }}/assets/js/charts.js?version=1ff0187"></script>
+		<script src="{{ site.baseurl }}/assets/js/charts.js?version=e7e9c5a"></script>
```
larsxschneider: Is there a way to generate a random hash or a hash over a file with Jekyll (via GitHub pages)?
/cc @parkr 😄

parkr: `site.github.build_revision` would work on Pages. https://help.github.com/articles/repository-metadata-on-github-pages/

pluehne: @parkr: Thanks for the pointer! This would generate new `?version=` hashes for each commit, right? Do you know of a way to, for example, get a hash based on the actual file content? In this way, we wouldn’t force client or server cache flushes with every single commit. (I’m just asking, your solution would solve our problem pretty well already!)

parkr: No hashes on a per-file basis are generated, so that’s the best solution that will achieve cache misses based on what’s in Pages.

pluehne: Okay, then this is the way to go. Thanks, @parkr!

pluehne:
In case someone is interested, I’ll come back to this soon. I wanted to have #81 in place before finishing this.

docs/spec/charts.js
```diff
@@ -1,4 +1,12 @@
-/* global createCollaborationChart, createHistoryChart, createList, createTable, createSpinner */
+/* global
+    aggregateTimeData,
+    createCollaborationChart,
+    createHistoryChart,
+    createList,
+    createTable,
+    createSpinner,
+    d3,
```
filmaj: Globals can be defined in a variety of ways in eslint. Other than inlining the list into particular .js files (like we do here), you can also specify them via eslint config files, similarly to how [we do in the browser-context specific `.eslintrc.json` file](https://github.com/Autodesk/hubble/blob/master/docs/assets/js/.eslintrc.json#L5-L13) (both `env` and `globals` config properties, in their own different ways, define what globals are expected to be available). I think we should move any globals defined by external libraries that our scripts depend on, like d3, jquery and moment.js, for example, to the configuration files, rather than as per-js-file running lists. In this way, if our webapp frontend ends up adding a new .js file, we won't have to duplicate the inline comment allowing for d3 globals.

pluehne: Thanks, that’s a good pointer. I wasn’t happy with inlining these globals either, and I second that it would be best to configure this in the `.eslintrc.json`.

I’ll return to this pull request in a couple of days. I’ll have to rebase these changes on top of the other new stuff, and then I’ll also adjust the global directives.

filmaj:
Huh, this code coverage comment uses the compact "header" layout. 🤔

docs/spec/charts.js
pluehne: @filmaj: I tried to change the [`globals` declaration for `d3`](https://github.com/Autodesk/hubble/blob/master/docs/assets/js/.eslintrc.json#L11) to `true` without success. I’d say let’s tidy up the global declaration in a separate pull request 🙂.

docs/assets/js/charts.js
```diff
@@ -142,6 +142,110 @@ function createSpinner(canvas)
...
+
+    // Turn date strings into proper date objects
+    for (let i = 0; i < data.length; i++)
+        data[i]['date'] = d3.isoParse(data[i]['date']);
+
+    // Sort data, just in case it isn’t already
+    data.sort((row1, row2) => row1['date'] - row2['date']);
+
+    const dateStart = data[0]['date'];
+    // Ranges are exclusive, so add one more day to include the last date
```
larsxschneider: `s/last date/last day/`?

docs/assets/js/charts.js
```diff
@@ -142,6 +142,110 @@ function createSpinner(canvas)
...
+
+    for (let i = 0; i < periods.length - 1; i++)
+    {
+        const t0 = periods[i];
+        const t1 = periods[i + 1];
+
+        // Note that this assumes complete data in the period
+        // Should data points be missing, aggregation methods such as the sum will lead to results that can't be
+        // compared to periods with complete data
+        // Hence, the maintainers of the data need to ensure that the input is well-formed
```
larsxschneider: micro-nit: It might be more readable if you add a `.` at the end of the sentence.

docs/assets/js/charts.js
```diff
@@ -142,6 +142,110 @@ function createSpinner(canvas)
...
+        const dates = data.filter(row => row['date'] >= t0 && row['date'] < t1);
+
+        let row = Object();
+        row['date'] = t0;
+
+        $.each(Object.keys(data[0]),
+            function(keyID, key)
+            {
+                if (key == 'date')
+                    return;
```
larsxschneider: micro-nit: that might be overly cautious as we use the value already in line 155 above

pluehne: This check is to exclude the dates themselves from being aggregated.

The data points all have the structure:

```
{
    "date": "yyyy-mm-dd",
    "key1": "value1",
    "key2": "value2",
    ...
}
```

We only want to aggregate the values of `key1` and `key2` but not `date`.

docs/assets/js/charts.js
pluehne: Refering to the “last date” was intentional, because this actually is about dates. However, “day” isn’t wrong either … not sure whether that’s worth rephrasing though.

docs/assets/js/charts.js
pluehne: I didn’t add periods, because we don’t have them in comments in other places. Should we do this everywhere then? Or put periods just *between* sentences in multisentence comments?

pluehne: Okay, then I’d say that we put periods *between* sentences systematically *except* for the end of comments (as is practiced by many print magazines, for example).

pluehne: I changed this as I suggested above.

docs/spec/charts.js
filmaj: Noted, I can take that on.

pluehne: That would be nice, thanks 😄!

filmaj: @pluehne I added another commit to my fork's branch of this PR that should fix that, feel free to pull in to this PR: https://github.com/filmaj/hubble/tree/patrick/improved-aggregation

pluehne: @filmaj: Thanks a lot! What I didn’t understand was that there are multiple `.eslintrc.json` files with different purposes—one for linting the assets themselves and one for linting the unit tests.

I squashed your changes into my commit above (to not have one commit do it improperly and another one fixing it right after), and I hope that this is fine with you 🙂.

filmaj: @pluehne yep perfectly fine 👍

PR #79, opened 2018-01-12
use ASCII only characters in scripts

larsxschneider:
Some Python version apparently cannot handle Unicode characters which
causes the following error:

```
    File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
       return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 881: ordinal not in range(128)
```

Fix this by using ASCII only characters.

Thanks to @rajivmucheli for patiently debugging the issue with me 😄


rajivmucheli:
Hi,

This issue popped up in v3.5.2.

Regards,
Rajiv

pluehne:
@rajivmucheli: Thanks for letting me know!

larsxschneider:
I have v.3.5.2, too, ... I guess his Python is compiled in a different way or something.

PR #80, opened 2018-01-15
Fix .gitignore path

pluehne:
This corrects the `.gitignore` path to the demo data directory, which was moved from `hubble-data` to `demo-data` earlier in e406ee5f10930f24367cbdc800362eeee907fa46.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,7 +1,7 @@
 _site
 
 # You can clone your "hubble-data" repository to this location for debugging
-hubble-data
+demo-data
 
 # nodejs bits
 node_modules


docs/.gitignore
```diff
@@ -1,7 +1,7 @@
 _site
 
 # You can clone your "hubble-data" repository to this location for debugging
-hubble-data
+demo-data
```
larsxschneider: Are you sure this is correct? Demo data should be under version control for the demo website!

pluehne: You’re right. `demo-data` should be removed from the `.gitignore` instead.

pluehne: For the record, I misunderstood the purpose of this line. It’s not for excluding the demo data directory (of course it isn’t, my mistake) but for excluding another local data directory for testing (I should have read the comment more carefully).

pluehne:
Let’s close this, as this was a misunderstanding 😄.

PR #81, opened 2018-01-15
Refactor DOM creation

pluehne:
This refactoring simplifies the HTML necessary to build charts by moving the creation of required wrappers to the JavaScript side.

This approach has multiple benefits.

1. First, the HTML part is more lightweight. Hence, future additions to the charts (such as action bars) will require much less changes to the HTML.
1. Second, we have more control over the layout on the JavaScript side without having to touch the HTML part.
1. Third, information concerning the chart content is grouped and separated from the layout part more nicely in this way.
1. Finally, with this preparation, we could render placeholders more nicely in the future, as they have a distinct style now.

The existing charts are adjusted to these changes and produce pixel-wise identical visualizations. Additionally, this unifies the way how charts, tables, and collaboration graphs are rendered.

### DIFF
--- a/docs/assets/css/stylesheet.css
+++ b/docs/assets/css/stylesheet.css
@@ -631,3 +631,8 @@ a
 		display: none;
 	}
 }
+
+.chart-placeholder
+{
+	display: none;
+}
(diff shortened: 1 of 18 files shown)


filmaj:
Would be good to update [`docs/README.md`'s Charts section](https://github.com/Autodesk/hubble/tree/master/docs#charts) to reflect these changes.

pluehne:
@filmaj: Good catch, I’ll do this later!

larsxschneider:
@filmaj Thanks for the review 👍 Do you generally agree with these changes? It's easier to review with this link: https://github.com/Autodesk/hubble/pull/81/files?w=1

PR #85, opened 2018-01-17
Add action bar with link to raw data to all charts

pluehne:
This adds an action bar to every chart with a button to download the raw data files in TSV format if existent. The action bar is designed to hold more actions in the future, most notably, a switch for the level of detail in the charts.

Along with the action bar, this commit introduces CSS for buttons.

Note that the action bar is to be considered as a “version 1” design to bring the functionality first, and the layout will probably change in a future release.

For reference, here is a screenshot with the new action bar highlighted:

action-bar

### DIFF
--- a/docs/assets/css/stylesheet.css
+++ b/docs/assets/css/stylesheet.css
@@ -636,3 +636,50 @@ a
 {
 	display: none;
 }
+
+.action-bar
+{
+	margin-top: 4rem;
+	margin-bottom: 1rem;
+}
+
+.action-bar h3
+{
+	margin: 0;
+}
+
+button,
+a.button
+{
+	display: inline-block;
+	padding: 0.35rem 0.8rem;
+	color: #606c71;
+	background: linear-gradient(-180deg, #fafafa, #f5f5f5);
+	outline: 0;
+	border: 1px solid #c4c4c4;
+	border-radius: 0.25rem;
+}
+
+button:hover,
+a.button:hover
+{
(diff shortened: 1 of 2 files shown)


filmaj:
Looks great... except on the Collaboration chart page, there is a big gap between the sidebars:

screen shot 2018-01-17 at 5 15 44 pm

Happens on both a recent Chrome and Firefox (desktop) version. Looks like it might be caused by the `col-aside` element in each row being vertically-aligned in the middle? Not sure...

pluehne:
Admittedly, this was intentional. I first put the action bar into the same `col-aside` as the other info boxes. However, this required some inelegant code (the action bar would have to be inserted *directly before the first* existing info box, but then it’s difficult to deal with charts that have *no* info boxes with descriptive texts).

The second reason for leaving the design like this is that I plan to implement a overhaul of the general layout. We’re starting to get to a stage where we have lots of UI elements that clutter the screen, and where it isn’t easy to quickly grasp what controls/elements belong to which group. That’s why I consider the action bar a “version 1” design (as described above) whose main intent is to bring the functionality.

@filmaj, @larsxschneider: Does that make sense to you? If not, I’m definitely open to adjust the layout for the time being.

filmaj:
Sounds fine to me, certainly not a blocker 👍

larsxschneider:
Looks good. #87 is minor nit that can be addressed in a separate PR.

Issue #82, opened 2018-01-15
Document + test data-config attribute for charts


filmaj:
It'd be nice to describe in the README what the `data-config` attribute customizes as well as what options exist. In a similar vein, formalizing tests for its functionality would be nice.

I'll be working on a PR for this, btw.

PR #86, opened 2018-01-17
Update readme to new chart interface

pluehne:
The HTML/JSON interface for creating charts was recently changed. This updates the readme accordingly. Additionally, this adds missing documentation about the allowed configuration options.


pluehne:
@filmaj: As you suggested, I updated the readme concerning charts to the new format that I introduced earlier.

Unfortunately, I just noticed that the configuration options part might interfere with your issue #82, where you said that you worked on something similar. I didn’t put too much effort into the changes above, so if you have a better idea how to describe the options, let’s take yours and just keep my description of the HTML structure 😃.

filmaj:
@pluehne I haven't started on those so I think we should what you have here.

docs/README.md
```diff
@@ -5,14 +5,58 @@ This web application is built as a [Jekyll](https://jekyllrb.com/) site and is i
 ### Charts
 
 Each `.html` page in this directory contains one or more charts.
-Each chart is defined as an `<svg>` element with a few special attributes that define their behavior:
+Charts are created from `<div>` containers of the class `chart-placeholder` by [`charts.js`](assets/js/charts.js).
```
filmaj: Perhaps we can word this as "Charts are dynamically created by [`charts.js`](assets/js/charts.js) from `<div class="chart-placeholder">` placeholders present in the HTML" ?

docs/README.md
```diff
@@ -5,14 +5,58 @@ This web application is built as a [Jekyll](https://jekyllrb.com/) site and is i
...
-1. `class`: the kind of chart that will be rendered (`chart`, `table`, or `collaboration`)
-1. `data-type`: the subtype of a chart (`history` or `list`, only applicable if `class` is set to `chart`)
+- `<h3>`: the title of the chart (required)
+- `<canvas>`: used to configure regular charts; change the tag to `<table>` for tables and `<svg>` for collaboration charts
+  - `data-url`: the URL to the `.tsv` file that shall be rendered by the chart (generated by the [updater](../updater/README.md))
+  - `data-type`: the chart type, supported types are:
+    - `history`: time series charts for tracking metrics over time
+    - `list`: horizontal bar charts for general numeric data
+    - `table`: tables directly rendered from the TSV source
+    - `collaboration`: chord diagrams used to visualize connections between entities
```
filmaj: While Hubble's sole chord diagram today is the Collaboration chart, perhaps it would align better to rename this chart type to 'chord'? This way if in the future we ship a _different_ kind of chord diagram, the type would apply a bit more generically?

pluehne: Good point, I’ll change the name to `chord`!

docs/README.md
pluehne: Thanks, I’ll incorporate your feedback!

pluehne:
@filmaj: I’ve addressed your two comments. I rebased your phrasing suggestion into the original commit, and I added a second commit for renaming the collaboration chart to *chord chart*.

filmaj:
Looks great!

PR #90, opened 2018-01-23
Refactor access to configuration options

pluehne:
This introduces two helper methods to check for configuration options and to read their values more concisely.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -113,11 +113,26 @@ const stackedBarChartDefaults =
     maintainAspectRatio: false
 };
 
+function hasConfig(element, key)
+{
+    return element.data('config') && (key in element.data('config'));
+}
+
+function readConfig(element, key)
+{
+    if (!hasConfig(element, key))
+        return undefined;
+
+    return element.data('config')[key];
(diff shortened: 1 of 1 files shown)


pluehne:
@filmaj: I can’t add you as a reviewer, but please let me know if you have feedback on this one 🙂—for instance, if you know a way to implement this even more compactly or without the duplicate checks.

pluehne:
@filmaj: ☝️ Seems like Codecov’s more compact “header” layout wasn’t applied here for some reason (also, the threshold of 2 % is not effective apparently). Could you check why that is?

filmaj:
Huh, weird. I will take a look. The comment layout is set as "header"... hrmm...

As for the threshold, I think that only works in the "negative" sense, that is, if code coverage _drops_ by 2%, then that will block the PR by setting the commit status to failed.

filmaj:
The other PRs with newer comments seem to be using the compact header format. Perhaps a hiccup?

pluehne:
@filmaj: Yeah, it seems to be working now for the newer pull requests. Let’s keep an eye on whether this surfaces again 🙂.

PR #91, opened 2018-01-23
Fix typo in chord chart code

pluehne:
The function “drawChord” was accidentally misspelled. I believe that it wasn’t *coordinates* that were meant but [*chords*](https://en.wikipedia.org/wiki/Chord_(geometry)), a term from geometry signifying line segments whose endpoints both reside on a circle.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -461,7 +461,7 @@ function createTable(table)
     });
 }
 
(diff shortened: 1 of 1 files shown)


pluehne:
@filmaj: ☝️ For some reason, the Codecov layout is fine with this pull request though 😄. I like it this way!

Issue #87, opened 2018-01-22
Disable "chart download" link for collaboration chart


larsxschneider:
In #85 we introduced an action bar with a download link for our charts. That's great for most charts but the collaboration chart has a raw data format that is not really useful as it cannot easily be processed in Excel and the like. That's why we should remove the download link for it.

pluehne:
Fixed by #92.

PR #92, opened 2018-01-23
Make the “raw data” button optional

pluehne:
This introduces a new configuration option “showRawDataLink,” through which the “raw data” button can be explicitly hidden (requested in #87).

This is used for hiding the “raw data” download button in the collaboration chart, where the raw data isn’t useful to view or edit in a spreadsheet application.

Also, the margins of chart titles were incorrect in the case that the corresponding action bar was hidden. This change applies the same style to all titles, regardless of whether an action bar is shown or not.

See how the “raw data” button is missing in the screenshot below 😄:

screenshot from 2018-01-23 14-19-05


larsxschneider:
Looks good! Minor nit: can you squash de83dbe38ba711f4869006c6518d8d889d4354ae and 118c075eeea9557346c39103db699b7ff263aad6 ?

pluehne:
@larsxschneider: Good idea, the commits belong together. I’ll squash them!

PR #93, opened 2018-01-23
Use GitHub build hash as parameter on assets

pluehne:
By using the GitHub Pages build hash as a “version” parameter on the asset URLs, we can force browser cache flushes of our JavaScript and CSS resources upon every commit.

While it wouldn’t be strictly necessary to change the version tag of a file on every commit (but only when it changes), this avoids having to manually change the version parameter after each change.

Note that I couldn’t test this yet, as I would have to replicate GitHub Pages somehow (maybe I could do this through a fork).

On another note, in the case that you deploy this locally, the variable `site.github.build_revision` will be rendered as an empty string. Not that that’s a problem from my point of view, but if you think this is something we should address, go ahead and tell me.

### DIFF
--- a/docs/_layouts/default.html
+++ b/docs/_layouts/default.html
@@ -5,29 +5,29 @@
 		<meta charset="UTF-8">
 		<title>Hubble Enterprise{% if page.title %}: {{ page.title }}{% endif %}</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1">
-		<link rel="icon" sizes="256x256" href="{{ site.baseurl }}/assets/icons/favicon-256x256.png?version=7c4fcab">
-		<link rel="icon" sizes="192x192" href="{{ site.baseurl }}/assets/icons/favicon-192x192.png?version=7c4fcab">
-		<link rel="icon" sizes="128x128" href="{{ site.baseurl }}/assets/icons/favicon-128x128.png?version=7c4fcab">
-		<link rel="icon" sizes="96x96" href="{{ site.baseurl }}/assets/icons/favicon-96x96.png?version=7c4fcab">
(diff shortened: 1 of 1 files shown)


docs/_layouts/default.html
```diff
@@ -5,29 +5,29 @@
...
 		<link rel="stylesheet" type="text/css" href="https://fonts.googleapis.com/css?family=Open+Sans:400,400i,700,700i" media="screen">
-		<link rel="stylesheet" type="text/css" href="{{ site.baseurl }}/assets/css/stylesheet.css?version=df12dbc" media="screen">
+		<link rel="stylesheet" type="text/css" href="{{ site.baseurl }}/assets/css/stylesheet.css?version={{ site.github.build_revision }}" media="screen">
 		<script src="{{ site.baseurl }}/assets/js/vendor/jquery-3.2.1.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/d3.v4.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/moment-with-locales.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/Chart-2.7.1.min.js"></script>
 		<script src="{{ site.baseurl }}/assets/js/vendor/spin-2.3.2.min.js"></script>
-		<script src="{{ site.baseurl }}/assets/js/charts.js?version=1ff0187"></script>
+		<script src="{{ site.baseurl }}/assets/js/charts.js?version={{ site.github.build_revision }}"></script>
```
larsxschneider: Is there any value to add this flag to the other (already versioned!) libraries? Maybe `moment-with-locales.min.js`?

larsxschneider: OK. That's an independent change. I'll merge is as-is.

PR #94, opened 2018-01-24
add organization activity chart

larsxschneider:
Calculate the number of active organizations for the last day, week,
and 4 weeks. Visualize the numbers in a new chart.

The naming is a bit inconsistent (`org` vs `orgs` vs `organization`). As always, I am open for suggestions!


updater/reports/ReportOrgActivity.py
```diff
@@ -0,0 +1,60 @@
...
+					users.id
+				FROM
+					repositories
+					JOIN users ON repositories.owner_id = users.id
+					JOIN pushes ON pushes.repository_id = repositories.id
+				WHERE
+					users.type = "Organization" AND
+					CAST(pushes.created_at AS DATE) BETWEEN "''' + str(timeRange[0]) + '''" AND "''' + str(timeRange[1]) + '''"'''
+
+		query +=   self.andExcludedEntities("users.login")
```
pluehne: The spacing after `+=` is odd here.

docs/demo-data/organization-activity.tsv
```diff
@@ -0,0 +1,7 @@
+date	last four weeks	last week	last day
+2018-01-23	26	23	18
+2018-01-22	26	22	16
+2018-01-21	26	21	13
+2018-01-20	26	22	16
+2018-01-19	26	23	17
+2018-01-18	26	23	12
```
pluehne: Maybe, we could add some more data points?

updater/reports/ReportOrgActivity.py
larsxschneider: Fixed!

docs/demo-data/organization-activity.tsv
larsxschneider: I would argue that this is enough to see how the chart would look like. The chart is not that unusual to warrant more data I would say 😄

pluehne: Okay, that was just a suggestion anyways 🙂.

updater/reports/ReportOrgActivity.py
pluehne: Thanks!

PR #95, opened 2018-01-25
add list of abandoned organizations

larsxschneider:
Find the organizations that have not received a push for the longest time.
Only look at organizations that have not received a push for at least one
year.

### DIFF
--- a/docs/housekeeping-abandoned-orgs.html
+++ b/docs/housekeeping-abandoned-orgs.html
@@ -1,0 +1,26 @@
+---
+layout: default
+title: Forks
+permalink: /housekeeping-abandoned-orgs
+---
+
+<div class="chart-placeholder">
+	<h3>Abandoned Organizations</h3>
+	<canvas
+		data-url="{{ site.dataURL }}/organizations-abandoned.tsv"
+		data-type="history"
+	></canvas>
+	<div class="info-box">
+		<p>
+			An organizations is considered <em>abandoned</em> if none of its repositories has received a push in the last year (ignoring <a href="https://help.github.com/articles/archiving-repositories/">archived</a> repositories).
+		</p>
+		<p>
+			If the content still has value, then you could <a href="https://help.github.com/articles/archiving-repositories/">archive</a> all repositories in these organizations to emphasize the fact that they are not maintained anymore.
+			If the content is not valuable anymore, then you could <a href="https://help.github.com/articles/deleting-an-organization-account/">delete the organization</a> to reduce clutter on your GitHub Enterprise appliance.
+		</p>
+	</div>
+</div>
+
+<div class="chart-placeholder">
+	<table data-url="{{ site.dataURL }}/organizations-abandoned-detailed.tsv"></table>
+</div>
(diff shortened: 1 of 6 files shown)


docs/housekeeping-abandoned-orgs.html
```diff
@@ -0,0 +1,32 @@
...
+
+<div class="chart-placeholder">
+	<h3>Abandoned Organizations</h3>
+	<canvas
+		data-url="{{ site.dataURL }}/organizations-abandoned.tsv"
+		data-type="history"
+	></canvas>
+	<div class="info-box">
+		<p>
+			An organizations is considered <em>abandoned</em> if none of its
```
pluehne: I’d suggest not wrapping these lines manually and leaving this to the editor program. My argument is that diffs tend to become very ugly when flow text is manually rewrapped later on.

docs/housekeeping-abandoned-orgs.html
```diff
@@ -0,0 +1,32 @@
...
+			repositories has received a push in the last year (ignoring
+			<a href="https://help.github.com/articles/archiving-repositories/">archived</a>
+			repositories).
+		</p>
+		<p>
+			If the content has still value, then you could <a href="https://help.github.com/articles/archiving-repositories/">archive</a>
```
pluehne: I’d write “still has value” instead of “has still value.”

updater/reports/ReportOrgsAbandoned.py
```diff
@@ -0,0 +1,38 @@
...
+			self.header = ["date", "abandoned organizations"]
+		self.data.append([str(self.yesterday()), len(self.detailedData)])
+		self.truncateData(self.timeRangeTotal())
+		self.sortDataByDate()
+
+	def query(self):
+		query = '''
+			SELECT
+				users.login AS "organization",
+				date(MAX(pushes.created_at)) AS "last push"
```
pluehne: Shouldn’t `DATE` be in uppercase for consistency?

updater/reports/ReportOrgsAbandoned.py
```diff
@@ -0,0 +1,38 @@
...
+			FROM
+				repositories
+				JOIN users ON repositories.owner_id = users.id
+				JOIN pushes ON pushes.repository_id = repositories.id
+			WHERE
+				    users.type = "organization"
```
pluehne: The indentation looks off by 3 spaces to me.

updater/reports/ReportOrgsAbandoned.py
```diff
@@ -0,0 +1,38 @@
...
+				AND repositories.maintained = 1 ''' + \
+				self.andExcludedEntities("users.login") + '''
+			GROUP BY
+				users.id
+			HAVING
+				MAX(pushes.created_at) < DATE_SUB(NOW(), INTERVAL 1 YEAR)
```
pluehne: I’d prefer using the Python-based approach to selecting time ranges instead of using `DATE_SUB(NOW())` and `INTERVAL 1 YEAR`.

The main reason for this is that I think it would be better to exclude today’s data from the report and to start with yesterday instead.

docs/housekeeping-abandoned-orgs.html
larsxschneider: true. but diffs for very long lines are ugly too. I dunno

larsxschneider:
@pluehne please review, again

PR #101, opened 2018-01-30
remove GitHub internal API requests from API statistics

larsxschneider:
GitHub seems to use its own API for internally. Filter these requests as
we are only interested in external API usage (because we can only control
external API usage).


larsxschneider:
We should mention in the release notes that this fix will lower the API request numbers a bit.

PR #104, opened 2018-01-31
report legacy admin teams

larsxschneider:
Organizations created before September 2015 might have legacy admin teams. In order to get the best performance out of GitHub Enterprise, these teams should be [migrated to the improved permissions model](https://help.github.com/enterprise/2.11/user/articles/migrating-your-previous-admin-teams-to-the-improved-organization-permissions/). Report the affected teams in a new dashboard. 

I recommend to review the PR commit by commit:
- the first commit contains an technically unrelated bug fix that I noticed during testing
- the second commit is preparation
- the third commit is the beef 😄


docs/_data/navigation.yml
```diff
@@ -50,4 +50,9 @@
     - title: "Forks"
       url: "/housekeeping-forks"
 - title: "Recommendations"
-  url: "/recommendations"
+  url: "/recommendations-tokenless-auth"
+  subnavigation:
+    - title: "Tokenless Authentication"
+      url: "/recommendations-tokenless-auth"
+    - title: "Legacy Teams"
+      url: "/recommendations-teams-legacy"
```
pluehne: I’d have named the URL `recommendations-legacy-teams`. I think that we wanted to have the navigation title (here, *Recommendations*) as the first component of the URL and have the rest match the subnavigation title (here, *Legacy Teams*).

Maybe, we should consider moving all the pages belonging to a navigation item into a subdirectory. In this way, the URLs would automatically be in the form `/recommendations/legacy-teams`, and this might make naming the URLs more consistent.

docs/assets/js/charts.js
```diff
@@ -520,7 +520,7 @@ function createTable(table)
 
                                 break;
                             default:
-                                cell.text(entry);
+                                cell.text(d);
```
pluehne: That makes sense, thanks for fixing this!

docs/recommendations-tokenless-auth.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: Recommendations
-permalink: /recommendations
+permalink: /recommendations-tokenless-auth
```
pluehne: Just a note not to forget renaming the permalink should we change it above 🙂.

updater/reports/ReportTeamsLegacy.py
```diff
@@ -0,0 +1,30 @@
+from .ReportDaily import *
+
+# Organizations created before September 2015 might have legacy admin
+# teams. In order to get the best performance out of GHE, these teams
+# should be migrated to the improved permissions model.
+# c.f.ghttps://help.github.com/enterprise/2.11/user/articles/migrating-your-previous-admin-teams-to-the-improved-organization-permissions/
```
pluehne: There seems to be an issue with `c.f.ghttps`. Looks like the `g` was intended to be a space character?

pluehne:
Wait, I think you forgot to add the HTML file 😅.

pluehne:
@larsxschneider: Wait, I think you forgot to add the HTML file 😅.

PR #105, opened 2018-01-31
Multiple views per chart and view switchers in action bar

pluehne:
This implements support for multiple detail views per chart (which can be configured through the `data-config` attribute in HTML) along with a handy view switcher in the action bar. Each view has its own configuration. For instance, there can be one view showing the daily data recorded in the last two months, while another view shows the weekly aggregated data over the last two years.

This pull requests exemplarily adds support for the view switcher to the *pull requests (total)* chart, while compatibility with the older configuration format is ensured.

I recommend reviewing this pull request commit by commit.

See the screenshots below to see how this looks in action 🙂.

screenshot from 2018-01-31 16-26-08

screenshot from 2018-01-31 16-26-03

screenshot from 2018-01-31 16-26-13


pluehne:
@larsxschneider: I just noticed that our reports truncate data after two years. Hence, the *all* view doesn’t really make sense, as it will show the same data as the *two years* view.

I’d suggest that we stop truncating data after 2 years, as we now have the option to show this period via the view switcher 😄.

However, that’s something we should address in another pull request.

PR #109, opened 2018-02-04
git-download: limit workaround for audit log roll bug

larsxschneider:
GHE 2.11.0 wrongly rolled the audit log only once a week. This problem
was fixed in 2.11.6. Limit the workaround to this version range.

### DIFF
--- a/updater/scripts/git-download.sh
+++ b/updater/scripts/git-download.sh
@@ -15,21 +15,25 @@ function ghe_greater_equal () {
 
 # Returns success (bash exit code 0) if the GHE version is in a given range.
 # The first value of the range is inclusive and last value exclusive.
-# E.g. 'ghe_between 2.11.4 2.12.2' means '2.11.4', '2.11.5', ..., '2.12.1'
-#      returns success and '2.12.2' does not return success.
+# For example, 'ghe_between 2.11.4 2.12.2' returns 0 (success) for the values
+# 2.11.4, 2.11.5, ..., and 2.12.1, while 1 (failure) is returned for 2.12.2.
 function ghe_between () {
     ghe_greater_equal "$1" && ! ghe_greater_equal "$2"
 }
 
-if ghe_greater_equal "2.11.0" ; then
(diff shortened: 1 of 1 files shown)


updater/scripts/git-download.sh
```diff
@@ -13,15 +13,27 @@ function ghe_greater_equal () {
...
-if ghe_greater_equal "2.11.0" ; then
-    # The "github-audit.log" log file introduced in GHE 2.11.0 is only rolled
-    # once a week. This was reported as a bug and is likely fixed in an
-    # upcoming version. In the meantime, we grep for all log entries in the two
-    # most recent log files (because the information from yesterday may or not
-    # be rotated already).
+# Returns success (bash exit code 0) if the GHE version is in a given range.
+# The first value of the range is inclusive and last value exclusive.
+# E.g. 'ghe_between 2.11.4 2.12.2' means '2.11.4', '2.11.5', ..., '2.12.1'
+#      returns success and '2.12.2' does not return success.
```
pluehne: Is a word missing here?

larsxschneider: ha. `s/returns/return/` ... does this lift confusion? I want to give an example what is included and what not. Do you have a suggestion to write this more clear?

pluehne: Okay—I thought the part `E.g. 'ghe_between 2.11.4 2.12.2' means '2.11.4', '2.11.5', ..., '2.12.1'` was a stand-alone sentence. Note that this would make sense, because `'ghe_between 2.11.4 2.12.2'` really stands for these values 😄.

Now I get what you want to say. The given call would return success for `'2.11.4', '2.11.5', ..., '2.12.1'` but not for `'2.12.2'`. I didn’t see that these values were the subject for the bottom part of the sentence.

I’d suggest writing the following to avoid ambiguities:

`For example, 'ghe_between 2.11.4 2.12.2' returns 0 (success) for the values 2.11.4, 2.11.5, ..., and 2.12.1, while 1 (failure) is returned for 2.12.2.`

PR #110, opened 2018-02-05
Move helper functions to separate module

pluehne:
I want to add a data repository scheme check to Hubble, which will share the data repository preparation steps with `update-stats.py`. For this reason, I thought it would be best to move these steps into a new helper module (see db13f474d9678957e6ed7d7ab46feedac4b778ce) so that they can be reused later on.

At the same time, I remembered that the `Report` module contained multiple helpers that don’t really belong there. For this reason, I moved them into the helper module as well (see c83d974f2e88598b506a67065f38af36085ed221).

I tested the changes and verified that everything works just as before.

### DIFF
--- a/updater/helpers.py
+++ b/updater/helpers.py
@@ -3,6 +3,8 @@ import os
 import subprocess
 import sys
 
+from config import *
+
 # Simple class to get floating-point values printed prettily
 class PrettyFloat(float):
 	def __str__(self):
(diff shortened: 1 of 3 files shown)


updater/helpers.py
```diff
@@ -24,3 +26,23 @@ def executeCommand(command, stdin = None, cwd = None):
...
+	return os.path.join(tmpDirectory, "hubble-data")
+
+# Create a local clone of the data repository if not present and update it
+def prepareDataDirectory(dataDirectory):
+	# Create data directory if not existing
+	if not os.path.exists(dataDirectory):
+		os.makedirs(dataDirectory)
+
+	# Get the data directory up-to-date
+	if not configuration["dryRun"]:
```
larsxschneider: That is side effect might surprise a caller of this function. I would move this check outside of the function.

pluehne: True, but even during a dry run we need the data directory to be created. This is because the data is still written to the TSV files but just not committed with Git.

Placing the check outside the function would prevent the data directory from being created if necessary, which would lead to errors.

As an alternative, we’d have to split the `prepareDataDirectory` function into two separate functions, one creating the directory if necessary and the other one updating the data to the latest version.

larsxschneider: Would it make sense to pass a parameter `dryrun` to the function?

larsxschneider: or call it `perform_clone`?

pluehne: Yep, that’s a good suggestion. I’ll make the change!

PR #111, opened 2018-02-05
Add forward compatibility check to updater

pluehne:
With this addition, the updater checks the schema version of the data in the data repository before attempting to run. In the case that the schema is outdated, an error is returned indicating that a migration
needs to be performed manually.

Additionally, the new script `check-schema.py` can be manually invoked to check the data schema prior to actually running the updater.


pluehne:
@larsxschneider: Ready for review 😄.

PR #112, opened 2018-02-05
add repo usage report

larsxschneider:
Calculate percentage of active repositories over the last four weeks
in organizations that have received pushes to at least one protected
branch and that have at least one topic assigned to them.

This way we can track if people use protected branches and topics!

---
Please review by commit. 1-3 are preparation, 4 is the beef 😄

### DIFF
--- a/docs/_data/navigation.yml
+++ b/docs/_data/navigation.yml
@@ -12,13 +12,15 @@
 - title: "Pull Requests"
   url: "/pr-usage"
   subnavigation:
-    - title: "Usage [%]"
+    - title: "Usage"
       url: "/pr-usage"
     - title: "Total"
       url: "/pr-total"
 - title: "Repositories"
-  url: "/repos-activity"
+  url: "/repos-feature-usage"
   subnavigation:
+    - title: "Feature Usage"
+      url: "/repos-feature-usage"
     - title: "Activity"
       url: "/repos-activity"
     - title: "Total"
--- a/docs/repos-activity.html
+++ b/docs/repos-activity.html
@@ -5,35 +5,57 @@ permalink: /repos-activity
 ---
 
 <div class="chart-placeholder">
-	<h3>Repositories (Active)</h3>
+	<h3>Active Repositories in Organizations</h3>
 	<canvas
 		data-url="{{ site.dataURL }}/repository-activity.tsv"
 		data-type="history"
 		data-config='{
 			"series": [
 				"in organizations (last four weeks)",
-				"in user accounts (last four weeks)",
 				"in organizations (last week)",
+				"in organizations (last day)"
+			],
+			"visibleSeries": [
+				"in organizations (last four weeks)"
+			]
+		}'
+		></canvas>
+	<div class="info-box">
+		<p>
+			Shows how many repositories are active in organizations.
+		</p>
+	</div>
+	<div class="info-box">
+		<p>
+			Repositories are counted as <em>active</em> if they received at least one push within the respective time period.
+		</p>
+	</div>
+</div>
+
+<div class="chart-placeholder">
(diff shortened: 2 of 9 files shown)


docs/repos-activity.html
```diff
@@ -5,29 +5,53 @@
...
+		}'
+		></canvas>
+	<div class="info-box">
+		<p>
+			Shows how many repositories are active in organizations.
+		</p>
+	</div>
+	<div class="info-box">
+		<p>
+			Repositories are counted as <em>active</em> if they received at
```
pluehne: Can you put this on one line?

docs/_data/navigation.yml
```diff
@@ -12,13 +12,15 @@
...
-    - title: "Usage [%]"
+    - title: "Usage"
       url: "/pr-usage"
     - title: "Total"
       url: "/pr-total"
 - title: "Repositories"
-  url: "/repos-activity"
+  url: "/repos-feature-usage"
   subnavigation:
+    - title: "Feature Usage"
```
pluehne: Good title for reports of this kind!

updater/reports/ReportRepoUsage.py
```diff
@@ -0,0 +1,53 @@
...
+		self.truncateData(self.timeRangeTotal())
+		self.sortDataByDate()
+
+	def query(self):
+		oneDayAgo = self.yesterday()
+		fourWeeksAgo = self.daysAgo(28)
+		query = '''
+			SELECT
+				"''' + str(oneDayAgo) + '''" AS date,
+				ROUND(100*COUNT(DISTINCT(protect.repository_id))/COUNT(DISTINCT(active.repository_id)))
```
pluehne: Can you put spaces around the arithmetic operators for consistency?

updater/reports/ReportRepoUsage.py
```diff
@@ -0,0 +1,53 @@
...
+					AS "protected branches [%]",
+				ROUND(100*COUNT(DISTINCT(topics.repository_id))/COUNT(DISTINCT(active.repository_id)))
```
pluehne: See above ☝️.

updater/reports/ReportRepoUsage.py
```diff
@@ -0,0 +1,53 @@
...
+					AS "topics [%]"
+			FROM
+				(
+				SELECT
+					repositories.id AS repository_id,
+					ref
```
pluehne: Could you write `repositories.ref` (if that’s the right one)? I had trouble seeing where `ref` comes from.

updater/reports/ReportRepoUsage.py
pluehne: Also, isn’t a float conversion necessary here to avoid an integer division?

pluehne: For the record, MySQL seems to perform a float division with `/` in all cases, while integer divisions require the `DIV` keyword. That’s strange, because other SQL implementations handle this differently as it seems …

updater/reports/ReportRepoUsage.py
```diff
@@ -0,0 +1,53 @@
...
+					AND users.type = "Organization" ''' + \
+					self.andExcludedEntities("users.login") + \
+					self.andExcludedEntities("repositories.name") + '''
+				GROUP BY
+					repositories.id
+				) AS active
+				LEFT JOIN (
+					SELECT repository_id, name
+					FROM protected_branches
+				) AS protect ON active.repository_id = protect.repository_id AND active.ref = "refs/heads/" + protect.name
```
pluehne: I’d abbreviate “protected Branches” as `protected` instead of `protect`, but that’s not really an issue.

PR #119, opened 2018-02-07
list of active repositories below the "Repository Feature Usage" graph

mlbright:
It is interesting to see what repositories are most active according to Hubble's definition. Moreover, it is very useful to have a complete list of all active repositories. Since this list can be big, display the top 50 on the webpage, and include a link to the complete data set.

### DIFF
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -554,14 +554,14 @@ function createTable(table)
                     return d;
                 });
 
-            let displayed = data;
+            let displayData = data;
             if (hasConfig($(table), 'slice'))
-                displayed = data.slice(readConfig($(table), 'slice')[0], readConfig($(table), 'slice')[1]);
+                displayData = data.slice(readConfig($(table), 'slice')[0], readConfig($(table), 'slice')[1]);
 
             let rows = d3.select(table)
                 .append('tbody')
                 .selectAll('tr')
-                .data(displayed)
+                .data(displayData)
                 .enter()
                 .append('tr');
 
--- a/docs/assets/js/charts.js
+++ b/docs/assets/js/charts.js
@@ -554,10 +554,14 @@ function createTable(table)
                     return d;
                 });
 
+            let displayed = data;
+            if (hasConfig($(table), 'slice'))
+                displayed = data.slice(readConfig($(table), 'slice')[0], readConfig($(table), 'slice')[1]);
+
             let rows = d3.select(table)
                 .append('tbody')
                 .selectAll('tr')
-                .data(data)
+                .data(displayed)
                 .enter()
                 .append('tr');
 
--- a/docs/repos-feature-usage.html
+++ b/docs/repos-feature-usage.html
@@ -31,6 +31,6 @@ permalink: /repos-feature-usage
 		data-config='{"slice": [0, 50]}'
 	></table>
 	<div class="info-box">
-		<p>The top 50 most active repositories, by number of pushes.</p>
+		<p>The top 50 most active repositories by number of pushes.</p>
 	</div>
 </div>
--- a/docs/repos-feature-usage.html
+++ b/docs/repos-feature-usage.html
@@ -23,3 +23,14 @@ permalink: /repos-feature-usage
 		</p>
 	</div>
 </div>
+
+<div class="chart-placeholder">
+	<h3>Most Active Repositories</h3>
+	<table
+		data-url="{{ site.dataURL }}/repository-usage-detailed.tsv"
+		data-config='{"slice": [0, 50]}'
+	></table>
+	<div class="info-box">
+		<p>The top 50 most active repositories, by number of pushes.</p>
+	</div>
+</div>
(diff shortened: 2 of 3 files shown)


larsxschneider:
You mention the top 50 truncation in the pull request body ... but this is not implemented yet, right?

mlbright:
Correct! The top 50 truncation is not implemented yet. Currently the entire list is displayed sorted by most number of pushes.

mlbright:
@larsxschneider I reworked the PR and I now filter the table to display only the top 50 most active repos.

docs/assets/js/charts.js
```diff
@@ -554,10 +554,14 @@ function createTable(table)
                     return d;
                 });
 
+            let displayed = data;
+            if (hasConfig($(table), 'slice'))
+                displayed = data.slice(readConfig($(table), 'slice')[0], readConfig($(table), 'slice')[1]);
```
larsxschneider: minor nits:
Maybe move this into a variable?
```
let tableSlice = readConfig($(table), 'slice');
``
Maybe rename `displayed` to `displayData`. `displayed` confused me because of the past tense.

@pluehne what do you think?

File: docs/repos-feature-usage.html [hunk 2]
```diff
@@ -23,3 +23,14 @@ <h3>Repository Feature Usage</h3>
...
 </div>
+
+<div class="chart-placeholder">
+	<h3>Most Active Repositories</h3>
+	<table
+		data-url="{{ site.dataURL }}/repository-usage-detailed.tsv"
+		data-config='{"slice": [0, 50]}'
+	></table>
+	<div class="info-box">
+		<p>The top 50 most active repositories, by number of pushes.</p>
```
larsxschneider: Is the `,` here correct?

docs/assets/js/charts.js
pluehne: I agree with @larsxschneider, `displayed` could be renamed to `displayData` for clarity what is actually displayed.

I believe that I have similar code with two `readConfig` calls as parameters on the same line somewhere else in the code ([here](https://github.com/mlbright/hubble/blob/ab2bf6c05d37289d69be2880d2d5e73fc9867b6d/docs/assets/js/charts.js#L454) and [here](https://github.com/mlbright/hubble/blob/ab2bf6c05d37289d69be2880d2d5e73fc9867b6d/docs/assets/js/charts.js#L559)), so I’ll take the blame for that 🙂. However, this might be the occasion to make the change at the other places as well.

pluehne: In this case, we should make the `tableSlice` constant. Also, I’d rather call it `configSlice` or something similar to show that it’s part of the configuration and not a slice of the table containing actual data.

updater/reports/ReportRepoUsage.py
```diff
@@ -13,6 +13,51 @@ def updateDailyData(self):
...
+			INNER JOIN
+			(
+				SELECT repository_id, COUNT(*) AS count
+				FROM pushes
+				GROUP BY repository_id
+			) AS push_tally
+			ON active.repository_id = push_tally.repository_id
+			ORDER BY count DESC
+		'''
+		return query
```
larsxschneider: Nit: In `activeReposQuery` below you assign `query` once. Here you append to the `query` variable. I would only use one stile. I think I prefer "assign once".

updater/reports/ReportRepoUsage.py
```diff
@@ -13,6 +13,51 @@ def updateDailyData(self):
...
+			FROM (
+		'''
+		query += self.activeReposQuery()
+		query += '''
+			) AS active
+			INNER JOIN
+			(
+				SELECT repository_id, COUNT(*) AS count
+				FROM pushes
+				GROUP BY repository_id
```
larsxschneider: Can't you just count the pushes in `activeReposQuery`? Then we wouldn't need the additional join here. I am not sure if I am missing something here...

docs/repos-feature-usage.html
pluehne: I’d say it’s a matter of preferences. Strictly, it isn’t necessary, but a comma can be used to separate additional information (in this case, “by number of pushes”) from the rest of the sentence. In this simple case, I’d leave the comma away.

updater/reports/ReportRepoUsage.py
pluehne: I find both styles okay as they are used here. Simple concatenations with `+` are fine for single variables etc., while entire lines (such as `self.activeReposQuery()` are broken down with `query +=`. I think it’s fine as it is 😄.

updater/reports/ReportRepoUsage.py
pluehne: @larsxschneider: I find both styles okay as they are used here. Simple concatenations with `+` are fine for single variables etc., while entire lines (such as `self.activeReposQuery()` are broken down with `query +=`. I think it’s fine as it is 😄.

docs/assets/js/charts.js
mlbright: I made the grammar change, but I'll leave the constant tableSlice changes for a separate PR.

larsxschneider:
superseded by #123

PR #124, opened 2018-02-14
Multiview collaboration chart

### LINKED ISSUE #115: Support multiple time ranges in the collaboration chart

pluehne:
This makes the collaboration chart multiview-capable and adds three views for different time ranges (2 months, 2 years, all data) to choose from.

screenshot from 2018-02-14 14-59-04

First, this extends the chord chart with support for multiple views. Users can switch between these with the view switchers that were previously implemented for history charts.

For this to work, the data file may contain multiple matrices in the same file (one after the other). If the configuration contains multiple entries in the “views” array, these will be mapped to the respective data matrices.

With this modification, the chord chart automatically generates a view switcher and an entity (for instance, organization) selector. The visualization is constantly kept in sync with these inputs.

Further, this adds three views to the collaboration chart along with some demo data for the live demo.

However, the demo data isn’t very interesting to look at and is subject to be replaced at a later point in time.

### To Do

- [ ] adjust the SQL queries to produce data in the new format (assigned to @larsxschneider)
- [ ] provide more interesting demo data

Closes #115.


pluehne:
@larsxschneider: When you adjust the updater report, you might want to think about whether the “two months, two years, all data” granularity makes sense here (haven’t thought much about this yet).

PR #127, opened 2018-02-26
Remove the “/day” suffix from data series labels

pluehne:
With the recent addition of multiple views with aggregated data, the `/day` label suffix doesn’t make sense in cases where weekly data is actually shown.

This removes the suffix from all labels to avoid confusion.

Note that Hubble will apply the header change upon the next run after applying the update. This means that between the time of applying the update and running Hubble, the web frontend will show empty charts, because the data series labels won’t match. However, I think that this is an acceptable caveat.

### DIFF
--- a/updater/reports/ReportGitDownload.py
+++ b/updater/reports/ReportGitDownload.py
@@ -12,10 +12,10 @@ class ReportGitDownload(ReportDaily):
 		if len(self.data) == 0:
 			self.header = [
 				"date",
-				"clones/day",
-				"fetches/day",
-				"clone traffic/day [GB]",
-				"fetch traffic/day [GB]",
+				"clones",
+				"fetches",
+				"clone traffic [GB]",
+				"fetch traffic [GB]",
 			]
 		self.data.append([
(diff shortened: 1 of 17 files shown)


pluehne:
@larsxschneider: I verified that the headers will be overwritten by Hubble as intended. The pull request is now ready to review :slightly_smiling_face:.

updater/reports/ReportGitDownload.py
```diff
@@ -12,10 +12,10 @@ def updateDailyData(self):
...
 			self.header = [
 				"date",
-				"clones/day",
-				"fetches/day",
-				"clone traffic/day [GB]",
-				"fetch traffic/day [GB]",
+				"clones",
+				"fetches",
+				"clone traffic [GB]",
+				"fetch traffic [GB]",
```
larsxschneider: The header is only rewritten if the `self.data` is empty. Since we reference the new header names in the JS code, the charts are empty 😢 

I wonder what's the best approach here. My gut feeling is to rewrite the header here. What do you think?

PR #135, opened 2018-02-28
Switch user activity chart to 4-week intervals

### LINKED ISSUE #129: Switch user activity chart to last 4 weeks (28 days) instead of month (30 days)
The [user activity chart](https://autodesk.github.io/hubble/users-activity) stems from before we came up with the idea to have all sliding windows and aggregation methods consider 4-week intervals instead of 30-day intervals. The benefit of this is that weekends are evened out, which avoids accidentally misinterpreting some fluctuations.

The user activity chart query should be changed, and the data format should be adapted to make things consistent.

pluehne:
The user activity chart stems from before most sliding windows and aggregation methods were switched from 30-day intervals to 4-week intervals. The benefit of this is that weekends are evened out, which avoids accidentally misinterpreting some fluctuations.

This changes the user activity chart accordingly for consistency.

Additionally, the time range computation in the updater report receives a minor refactoring based on the more recent reports.

Note that this change will invalidate previously recorded data based on 30-day intervals. However, no action is taken, because the difference between data aggregated over 28 and 30 days is expected to be subtle.

Resolves #129.

### DIFF
--- a/docs/demo-data/users.tsv
+++ b/docs/demo-data/users.tsv
@@ -1,4 +1,4 @@
-date	pushing commits (last day)	pushing commits (last week)	pushing commits (last month)	using license
+date	pushing commits (last day)	pushing commits (last week)	pushing commits (last four weeks)	using license
 2017-09-25	104	184	235	330
 2017-09-24	13	182	234	326
 2017-09-23	12	181	238	325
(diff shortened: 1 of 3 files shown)


pluehne:
@larsxschneider: I rebased this on top of the merged changes of #137 and #138.

Issue #139, opened 2018-03-21
Git traffic and tokenless authentication charts empty after GHE upgrade


talktopiyush:
Hello there,

We have hubble as service on a dedicated machine and we recently did GHE upgrade from v2.11 to v2.12.6 and after this hubble does not show the git traffic/ there is no data in git-download.tsv

fyi, there are no errors reported when we run `update-stats.py`.

In general do we need to do anything before or after we apply an update to GHE.

Regards
Piyush

talktopiyush:
similarly no data (or marked as zero against date) in tokenless-authentication.tsv since the upgrade.

pluehne:
Hi @talktopiyush! Thanks for reaching out about these two issues. We noticed the issues with the *Git traffic* charts as well, and they should be fixed by #138.

I’d like you to check out the latest commit in the `master` branch and to run the Hubble updater again. This should resolve the first part of the issue. (The context is that we made changes to the naming of the data series, and unfortunately, these weren’t reflected in the data files that the updater produces. Hence, the dashboard couldn’t find the data anymore and would show empty charts.)

Concerning the *tokenless authentication* chart, there was a change in GitHub’s log file format from 2.11 to 2.12 that broke our script for that chart’s data. I’m currently working on this and hope to have a patch out later today.

As a general suggestion, we recommend that you use the latest release (currently, 0.2.0) and not the `master` branch directly. Also, we’re trying to avoid issues on `master` as best as we could, but we can’t guarantee forward compatibility with new GitHub Enterprise versions unfortunately.

pluehne:
@talktopiyush: In the meantime, please let me know whether #138 fixed the issues you experienced with the *Git traffic* charts.

pluehne:
@talktopiyush: I forgot to notify you, but we merged my fix #141 that should resolve your issue with the *tokenless authentication* chart. It would be great if you could test all these changes and tell me whether we can close this issue :smiley:.

rajivmucheli:
Hi @pluehne post upgrading, i can confirm Git Traffic Chart is populating data but tokenless-authentication-detailed.tsv FAILS,  please note tokenless authentication.tsv gets updated.

pluehne:
@rajivmucheli: Happy to hear that the *Git traffic* chart is fixed for you now. Can you explain what exactly fails with the detailed *tokenless authentications* report? Is it just empty or do you get an error message?

[...]

PR #141, opened 2018-03-21
Fix tokenless authentication chart

pluehne:
The format of the `gitauth.log` file changed from GitHub Enterprise 2.11 to 2.12. This makes it necessary to adjust the regular expression responsible for fetching user names and repositories.

Unfortunately, it seems that we can no longer rely on the order of the fields in gitauth.log. For this reason, multiple grep iterations are performed and the regular expression is changed to use look-aheads to allow for arbitrary field orders.

For backward compatibility with GitHub Enterprise 2.11 and earlier, a switch is implemented that executes the old query if necessary.

Addresses the second issue mentioned in #139.

### DIFF
--- a/updater/scripts/tokenless-auth.sh
+++ b/updater/scripts/tokenless-auth.sh
@@ -3,11 +3,36 @@
 # Lists users that make the most requests with username/password
 # c.f. https://help.github.com/enterprise/2.11/admin/guides/user-management/using-ldap/#disabling-password-authentication-for-git-operations
 #
+
+function ghe_greater_equal () {
+    cat /etc/github/enterprise-release |
+        perl -sne '
+            use version;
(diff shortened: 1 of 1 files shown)


pluehne:
@larsxschneider: Please have a look at my changes. I hope that you’re okay with the look-ahead solution. What it does is capture the `member` and `path` fields, but with a look-ahead, which means that the cursor position is not altered, which is why the fields are matched in arbitrary orders.

I’m still thinking whether we could simply support GitHub Enterprise 2.11 by extending the `status` field to match either `200` or `OK` and the `hashed_token` field to match either `nil` or to be missing entirely. What do you think?

pluehne:
@larsxschneider: I implemented the backward compatibility check, and this pull request is now ready for review :smile:.

PR #143, opened 2018-03-27
Improve tooltips when aggregating data

### LINKED ISSUE #130: Have tooltip show date ranges for aggregated data
Sometimes, when looking at weekly-aggregated charts, it confuses me to see the date of the first day of the in the tooltips when hovering data points. Users could also be confused, because it’s not clear whether the shown value represents a single date or an entire week (and where the week starts and ends is also undocumented).

This could be addressed by having Chart.js show the respective date range in the tooltips. For data series that have been aggregated by `first` and `last` (which select only the chronologically first/last within each period), the tooltip should still show the respective (single) date. However, for data aggregated with `sum`, `mean`, `med`, `min`, and `max`, we could show the time range instead.

pluehne:
This implements multiple improvements to the tooltips that are shown when hovering data points in views displaying aggregated data.

First, the tooltips now show the date range of the period of time over which the data has been aggregated (and not just the first date of the range). An exception to this are the `first` and `last` aggregation methods (which only pick the first or last value from the available time range), where the respective date is shown instead.

For reasons of clarity, the date format was changed to show abbreviated month names instead of full ones.

Second, the applied aggregation method is shown in parentheses next to the date range. This makes it easier to understand how to interpret the data. Again, this excludes the `first` and `last` aggregation methods, where the date itself already provides all the information needed.

Implementation-wise, new auxiliary columns were introduced. These are distinguished from regular columns with a leading underscore character and subsequently ignored from the chart rendering. This mechanism is useful for annotating data, currently with information for showing the tooltips.

I also updated Moment.js to 2.21.0 because of an issue with how UTC timestamps are handled.

Closes #130.

---

screenshot-weekly-sum

---

screenshot-weekly-first

---

screenshot-daily


docs/assets/js/charts.js
```diff
@@ -63,7 +63,7 @@ const timeSeriesChartDefaults =
                 time:
                 {
                     format: 'YYYY-MM-DD',
-                    tooltipFormat: 'D MMMM YYYY',
+                    tooltipFormat: 'D MMM YYYY',
```
larsxschneider: Where is `format` used? I wonder if the formats should match...

pluehne: The `format` is the default tooltip configuration if you haven’t set a tooltip callback.

Having it spelled out here is mostly a defensive measure now that we have a callback. But I agree that it’s not really necessary anymore. Depends mostly on whether we can reuse that default configuration at a later point in time or not.

docs/assets/js/charts.js
```diff
@@ -382,14 +402,51 @@ function createHistoryChart(canvas, actionBar)
...
+                    // In the history chart, tooltips are rendered for individual data points.
+                    // Hence, the tooltipItem array contains exactly one element
+                    const dataPoint = data.datasets[tooltipItem[0].datasetIndex].data[tooltipItem[0].index];
+
+                    const date = dataPoint.x;
+                    const dateRange = dataPoint.dateRange;
+
+                    const suffix = (dataPoint.infoText === undefined)
+                        ? ''
+                        : ' (' + dataPoint.infoText + ')';
```
larsxschneider: Nit: Wouldn't is be more straight forward to set `row['_infoText'] = ''`? Then you don't need this condition.

docs/assets/js/charts.js
```diff
@@ -382,14 +402,51 @@ function createHistoryChart(canvas, actionBar)
...
+
+                    if (dateRange === undefined)
+                        return moment(date).utc().format('D MMM YYYY') + suffix;
```
larsxschneider: Would it make sense to define the tooltip format as some kind of global constant? We replicate it 5 times in this PR.

docs/assets/js/charts.js
pluehne: I thought I’d rather decouple the presentation (putting the info text in parentheses after the title) from the data itself. This keeps the data array rather clean I think even if this conditional is the consequence.

docs/assets/js/charts.js
pluehne: I also thought about this. However, the interface of Moments.js doesn’t look like this could be extracted.

I would have expected Moments.js to work like this:

```
moment.utc().format('D MMM YYYY')(1234567890);
```

This would have enabled us to extract the formatting function like this:

```
const formatTime = moment.utc().format('D MMM YYYY');
formatTime(1234567890);
```

However, as you can see, the interface isn’t like this. Yes, we could move the format into a global configuration, but then there’s still the UTC part that’s replicated everywhere.

pluehne: But now thinking of this, let me have a quick look whether there is another interface that would do the trick.

pluehne: We could also simply write a wrapper function:

```js
function formatTime(time)
{
    return moment(time).utc().format('D MMM YYYY');
}
```

Would you prefer it in this way?

pluehne:
@larsxschneider: I addressed your feedback concerning the code duplication with date formatting in 1db89be601b81b29ed6ad6f7c41a40ec35f329b7. Please have a look at the changes :slightly_smiling_face:.

PR #146, opened 2018-03-29
added meta.tsv, updated documentation, fixed calendar import

IngoS11:
added an example meta.tsv file that can be used for the hubble-data repository. Update the README.md to tell people that a meta.tsv file has to be present inside hubble-data. Added an Exception and print output in case there is no meta.tsv file. Fixed missing import statement in ReportPRUsage.py for calendar

### DIFF
--- a/updater/schema.py
+++ b/updater/schema.py
@@ -20,9 +20,11 @@ def checkSchemaVersion(dataDirectory):
 				if row[0] == "schema-version":
 					schemaVersionLocal = int(row[1])
 					break
-	except IOException as e:
+	except IOError:
 		print("error: the data repository has no file named meta.tsv", file = sys.stderr)
 		sys.exit(1)
+	except:
+		pass
 
 	if schemaVersionLocal < schemaVersion:
 		print("error: the data repository has an outdated scheme and needs to be migrated", file = sys.stderr)
--- a/updater/schema.py
+++ b/updater/schema.py
@@ -20,8 +20,9 @@ def checkSchemaVersion(dataDirectory):
 				if row[0] == "schema-version":
 					schemaVersionLocal = int(row[1])
 					break
-	except:
-		pass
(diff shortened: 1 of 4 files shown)


pluehne:
@IngoS11: Thanks for these patches!

I believe that it would be best to handle empty `meta.tsv` files within the `update-stats.py`. This would make the users’ lives easier, because we wouldn’t need to add more steps to the documentation.

Aside from that, I’ll look into the `import calendar` issue. To that end, could you pleaase tell me what Python version you’re using?

updater/schema.py
```diff
@@ -20,6 +20,9 @@ def checkSchemaVersion(dataDirectory):
 				if row[0] == "schema-version":
 					schemaVersionLocal = int(row[1])
 					break
+	except IOError:
+		print("error: the data repository has no file named meta.tsv", file = sys.stderr)
+		sys.exit(1)
```
larsxschneider: 👋 @IngoS11 ! Thanks for fixing this issue!

I wonder, couldn't we just assume the default `schemaVersion` in case of an `IOError`? That means we would set:
```
schemaVersionLocal = schemaVersion
```
here.

What do you think?

pluehne:
@IngoS11: Thanks for your contributions. I think that it’s nicer to add support for empty `meta.tsv` files to the updater than to force users to create one by themselves (which introduces another vector of potential issues).

For this reason, I decided to split your pull request into the `calendar` import fix (#152) and the added support for empty `meta.tsv` files (#153).

Again, thanks for helping us out, and I hope that you’re fine with me closing this pull request, now that it’s superseded by #152 and #153.

Just on a side note, feel free to open separate pull requests even for minor things such as the `calendar` import—this also allows for merging and reviewing this more quickly :slightly_smiling_face:.

pluehne:
Superseded by #152 and #153.

PR #148, opened 2018-04-05
Add API Requests By User

dfarr:
This adds a report to collect the top 20 API users each day, and
log the total number of requests each hour. Reports this info in
a simple table, this is handy for identifying users who are
approaching, or over the API rate limit.

An example of the API Rate Limit page:
screen shot 2018-04-05 at 3 38 40 pm


larsxschneider:
👋 @dfarr ! I have a few questions:

(1) Was there a specific reason to print the numbers per hour? I am asking because the finest granularity for all other numbers in Hubble is a "day" and I wonder if we should stick to that.

(2) Would it make sense to move your "Top API Users" into the [API requests pages](https://autodesk.github.io/hubble/housekeeping-api-requests)?

 Thanks for this PR, this looks really useful 👍 😊

dfarr:
I think it would make sense to do top API users - does regenerating this every day make sense? If so it would be a pretty simple tweak to the script and simply moving the table to a different page :)

toddocon:
I think generating this report every day makes sense. I have some very heavy API (and polling) users and would like to make this information more visible.

larsxschneider:
Yeah, top API users per day makes total sense. @dfarr Do you want to tackle this?

@toddocon AFAIK this should work with and without rate limiting enabled.

dfarr:
Will do :+1:

From my understanding, the `api-requests-by-user.sh` script will extract the data regardless of whether or not rate limiting is enabled. Rate limiting information does surface in some fields of the unicorn logs, but these fields are ignored by this script.

dfarr:
Updated! Here is a screenshot of the updated housekeeping/api-requests page:

screen shot 2018-04-16 at 1 38 59 pm

toddocon:
I like it! Can't wait to get this in my version.

docs/housekeeping-api-requests.html
```diff
@@ -50,6 +50,20 @@ <h3>API Requests</h3>
 	</div>
 </div>
 
+<div class="chart-placeholder">
+	<h3>Top API Requests by Users</h3>
```
pluehne: I’d drop the “Top” in the title, because it’s common for our charts only to show the top entries (for instance, in the [Git traffic table](https://autodesk.github.io/hubble/housekeeping-git-traffic)).

larsxschneider:
@dfarr Thank you very much 😃 👍 

I rebased your PR and made a few minor modifications. Please take a look: https://github.com/Autodesk/hubble/pull/156/commits/f14daab05f8981987f123857de4f9365531dc6d2

Issue #147, opened 2018-03-29
Date mismatch—timezone issue?


toddocon:
Hubble data is collected from the prior day's rotated logs. That data is collecting properly on my system and being pushed to GHE. 

When I view the Hubble graphs, though, the date that is shown is day -2 but the actual data is from day -1 (which is expected). I reside in the Mountain Time Zone. If I Remote Desktop into a system in Pacific Time Zone the date that is shown is day -1 (which is expected) and the data is still OK. 

So for me, I'm seeing a display date issue but the underlying data is OK. 

I am using the latest released version of Hubble (0.2.0?) and GHE 2.10.6. There was no date display issue until I updated to Hubble 0.2.0 so maybe something was introduced in the new release. 

Our 2.12.X upgrade is a few weeks out. I need to install this Hubble on stage to see if the problem is still there.

pluehne:
Hi @toddocon! I recently did some changes to the timezone handling (#143), so it could well be related to that. When you said that you use the latest version of Hubble, did you mean the `v0.2.0` release or a more recent version from `master`?

toddocon:
Hi Patrick

I am using the latest released Hubble, v0.2.0. We are upgrading to 2.12.8 in a few weeks if that matters.

pluehne:
@toddocon: Can you reproduce this issue on [autodesk.github.io/hubble](https://autodesk.github.io/hubble/)? That page is running an updated version of the dashboard with my changes in #143.

pluehne:
@toddocon: It’s not urgent, but could you test whether the date mismatch issue is still present when updating to the newest `master`? I’d like to know whether there is still some bug to be chased :slightly_smiling_face:.

toddocon:
@pluehne We upgraded to GHE 2.12.9 on Friday and I updated to the latest Hubble master as well. I am happy to report that today (4.16), hovering over a data bubble shows the date as 4.15. So the fix worked! All is well now. 

Thanks and regards.

pluehne:
@toddocon: That’s great to hear! That means that we can close this one and #118?

toddocon:
@pluehne : Yes, we can close both issues. Thanks for the help!

PR #149, opened 2018-04-09
Fix incorrect dates shown at ends of time ranges

pluehne:
Because of a mistake, the end date of time ranges as shown in the dashboard was not copied by value but by reference. As an effect, displayed dates would be repeatedly decremented by one second every time users hovered data points. For this reason, displayed dates could vary when hovering them multiple times.

This is easily fixed by copying the affected variable by value.


pluehne:
@toddocon: I discovered this issue while debugging your bug report in #147. I think this one could be related or even the explanation for #147, but I’m not entirely sure. I manually set my time zone to PST, and all dates look fine to me now.

So if you could try out this patch, this would be much appreciated :smile:.

docs/assets/js/charts.js
```diff
@@ -429,7 +429,7 @@ function createHistoryChart(canvas, actionBar)
 
                     // The date range is of the form [t0, t1), inclusive of t0 but exclusive of t1.
                     // Hence, subtract one second from t1 to obtain the previous date in UTC
-                    let t1 = dateRange[1];
```
larsxschneider: Do you see an easy way to cover that with a test case? E.g. move `tooltipTitleCallback` to a "real" function and then test this function with some demo data?

pluehne: I thought about this as well, but now that you think of unit-testing this as well, I probably should do this tomorrow :slightly_smiling_face:.

pluehne:
@larsxschneider: I added some unit tests, so please review this again :slightly_smiling_face:.

PR #155, opened 2018-04-23
Improve some descriptions

larsxschneider:


docs/housekeeping-git-traffic.html
```diff
@@ -120,7 +120,7 @@ <h3>Transferred Data</h3>
 		}'></canvas>
 	<div class="info-box">
 		<p>
-			The amount of data transferred for clones and fetches.
+			The amount of data transferred for clones and fetches (<b>not</b> including <a href="https://git-lfs.github.com/">Git LFS</a> traffic).
```
pluehne: I’d `<em>` this instead of making the text bold. (Note that I generally prefer `<em>` over `<i>` and `<strong>` over `<b>`, because `<i>` and `<b>` imply a certain style and not structure, which is something CSS should take care of and not HTML.)

docs/repos-activity.html
```diff
@@ -91,7 +91,7 @@ <h3>Most Active Repositories</h3>
 		data-type="table"
 	></table>
 	<div class="info-box">
-		<p>The top 50 most active repositories by number of pushes.</p>
+		<p>The top 50 most active repositories by number of pushes over the last 4 weeks. The number of pushers and pushes are summed up over the last 4 weeks.</p>
```
pluehne: Can you add a line break between the two sentences? I’d format is as follows:

```html
<p>
	The top 50 most active repositories by number of pushes over the last 4 weeks.
	The number of pushers and pushes are summed up over the last 4 weeks.
</p>
```

pluehne: I find these two sentences somewhat confusing. Why does the first sentence mention *pushes* but the second one *pushers and pushes*. Couldn’t both be explained in one sentence? Also, why is the aggregation period “over the last 4 weeks” repeated twice?

larsxschneider: > Why does the first sentence mention pushes but the second one pushers and pushes. 
The first sentence is supposed to explain that the table is sorted by number of pushes per repository over the last 4 week

> Also, why is the aggregation period “over the last 4 weeks” repeated twice
Yeah, that's not nice.

How about this:
The top 50 most active repositories sorted by number of pushes per repository.			The number of pushes and pushers are summed up over the last 4 weeks.

pluehne: Can you make the latter “number” plural? Else, one might think that the number comprises both pushes and pushers (a metric that doesn’t really make sense). I’d also add an article in the first sentence.

How about this:

> The top 50 most active repositories sorted by the number of pushes per repository.
> The numbers of pushes and pushers are summed up over the last 4 weeks.

I’m also wondering whether we should throw in a “respectively” in the second sentence before ”last 4 weeks” to make clear that the sum is over the 4 weeks before the data point and not before today.

PR #161, opened 2018-07-22
track GHE internal API requests

larsxschneider:
GHE uses its API for internal requests as well. Previously we ignored
these calls. Let's track them too as they could have an impact on the
performance as well.


larsxschneider:
This only tracks internal API requests that go via haproxy... that's not useful.

PR #166, opened 2018-10-01
Decompose charts.js

pluehne:
The JavaScript part of the dashboard is contained in a single [`charts.js`](https://github.com/Autodesk/hubble/blob/f542611d0f8af2ff2ef69500c5680d5ca2ebfc49/docs/assets/js/charts.js) file that has grown over time and that became quite hard to read, maintain, or extend.

This refactoring uses the `include_relative` Jekyll directive, which can safely be used on GitHub Pages, because it is part of the `gh_pages` gem. This directive allows assets (not just pages) to be composed of partial source files.

This commit splits `charts.js` into special-purpose partial JavaScript files containing shared utilities, style-related definitions, as well as the charts themselves.

Unfortunately, this approach breaks our testing setup, the reason being that the testing framework looks at the unprocessed source files, and the `include_relative` directives aren’t valid JavaScript code. To work around this issue, this commit first invokes Jekyll to generate the final `charts.js` from the partials before running the tests.

---

Note that I didn’t touch any code that I moved from `charts.js` to the partials (not even whitespace). This pull request is only concerned with decomposing the JavaScript part of Hubble. While some parts of the code may be improved in one way or another, I’ll leave this to future pull requests.


pluehne:
@larsxschneider: I appeased Travis (I was missing a package and Travis’s bash apparently didn’t understand `{}` expansion), and the tests are running fine now. Would be cool if you could give this pull request a look now :slightly_smiling_face:.

PR #173, opened 2018-11-28
Fix unit tests for time data aggregation

pluehne:
By accident, these unit tests used the assignment operator `=` instead of the comparison operator `==`. This patch remedies the issue and, at the same time, switches to the more idiomatic `toEqual` method.


pluehne:
[The Travis tests passed](https://travis-ci.org/Autodesk/hubble/builds/460864589), but for some reason, the status checks aren’t updated and turned green.

Anyway, merging this now.

PR #174, opened 2018-11-29
Add recommended Git versions chart

pluehne:
This adds a chart that shows how many users are using Git clients that are recommended, outdated, and vulnerable as a stacked area chart.

### Screenshot
screenshot from 2018-11-29 13-46-47

### Details

A Git version is considered _vulnerable_ at a specific date if there existed a vulnerability that was publicly disclosed at that time. A client is counted as _outdated_ if there was a newer patch release available at that date. All Git clients that are neither vulnerable nor outdated by these definitions are _recommended_ for productive use.

The chart obtains its data from the previously introduced Git release database. The `GitVersionsDatabase` class reads this database and flags Git versions according to the recommended/outdated/vulnerable classification. For this to work, Git versions are compared against the affected version ranges stated in the Git version database according to the semantic versioning specification.

For visualizing the results, a new area chart is implemented. The actual recommended Git versions chart is built on top of the area chart and performs only the operations necessary to prepare the data accordingly.

I manually crafted demo data to visualize the chart on autodesk.github.io/hubble and for the unit tests to work.

Unit tests cover parsing and comparing semantic versions, the correct annotation of recommended, outdated, and vulnerable versions in the database, and the correct computation of how many versions were recommended, outdated, and vulnerable at a given date based on recorded usage data.

### Neat Features

Stacked **area charts** are now available (a7318668cae0876ea72b4aea38fae8921dd74025) and include customization options such as **whether or not to normalize the data** (14f95b8ebb309900dfa10dc75f7ffb396c913bde) and **custom color selection** for the datasets (516239dce91c48ea0724da5d102aed89418c3306).

The Git release database is **preferentially retrieved from autodesk.github.io/hubble** if accessible to obtain more recent data (d73269a9f3e3b50ca53c1bd52649c32891832717). If the online database file can’t be reached, the local copy will be used as a fallback. I also added a **script to easily update the Git release database** in the future (e8f00fadd813e69610d3fbbb11cdb345a01e29e7). All we need to do is invoke this script and publish these changes on the `master` branch, and the most recent Git version information will immediately be available to all Hubble users.

### Notes

I spent a lot of time dissecting this huge pull request into easy-to-review commits. So this pull request is best reviewed commit by commit.

This implementation leverages many ECMAScript 6 features that we didn’t use before (classes, for-of loops, etc.) and that are widely supported by browsers. While I know that this introduces inconsistencies with the existing code base, I think that this change is for the better. I’ll go over the existing chart implementations and adjust them in analogy to what I implemented for area charts and the _recommended Git versions_ chart.


docs/tools/update-git-versions.json.sh
```diff
@@ -0,0 +1,53 @@
...
+# anyway by now. Release candidates are also removed.
+TZ=UTC git -C git tag -l 'v*' --format='%(refname:short)	%(taggerdate:iso-local)' --sort=-taggerdate \
+	| grep -v '^v[01]' \
+	| grep -v 'rc' \
+	| perl -pe 's/(\d{4,}-\d{2,}-\d{2,}) \d{2,}:\d{2,}:\d{2,} \+0000/\1/g' \
+	| cut -c 2- \
+	> git-releases.tsv
+
+# Translate the TSV file to JSON.
+jq --slurp --raw-input --raw-output --tab \
```
larsxschneider: Minor nit: `jq` is not available everywhere. You could check if it is installed like this: https://github.com/Autodesk/enterprise-config-for-git/blob/4d592a7f18dd6328f278ca8cdf1f6a81c7411fb1/lib/setup_helpers.sh#L48-L50

docs/tools/update-git-versions.json.sh
```diff
@@ -0,0 +1,53 @@
...
+git_versions_file=$1
+
+if [[ -z $git_versions_file ]]
+then
+	echo "Usage: update-git-versions.json.sh <git-versions.json file> [--in-place]"
+	exit 1
+fi
+
+# Clone a fresh repository
+if [[ ! -d git ]]
```
larsxschneider: Maybe this directory should be explicitly ignored in `.gitignore` of the root repo to avoid confusion? Or clone it to `/tmp`?

docs/assets/js/_partials/semantic-versioning.js
```diff
@@ -0,0 +1,94 @@
...
+            case '=':
+            {
+                if (comparisonResult != 0)
+                    return false;
+                break;
+            }
+            case '<':
+            {
+                if (comparisonResult >= 0)
+                    return false;
```
larsxschneider: Minor nit: Maybe something like that would be easier to read (same for the comparisons below)?
```
if (!(comparisonResult < 0))
                    return false;
```

docs/assets/js/_partials/git-versions-database.js
```diff
@@ -0,0 +1,110 @@
...
+    load(callback, customDataURL = undefined)
+    {
+        // If the data has already been collected, immediately return it
+        if (this.data != undefined)
+        {
+            callback(null, this.data);
+            return;
+        }
+
+        const dataURL = (customDataURL != undefined) ? customDataURL : GitVersionsDatabase.dataURL;
```
larsxschneider: I wonder if we should let the Hubble user activate the call to github.com here explicitly?

docs/assets/js/_partials/chart-area.js
```diff
@@ -0,0 +1,251 @@
...
+            this._aggregateData(view);
+
+            // Build the actual data used by Chart.js for rendering
+            this._buildChartData(view);
+        }
+
+        this.chartJSChart = new Chart(
+            this.context,
+            {
+                type: 'line',
```
larsxschneider: Interesting that the type is still `line` although it's an area chart.

docs/assets/js/_partials/chart-area.js
```diff
@@ -0,0 +1,251 @@
...
+
+    _initializeViews(data)
+    {
+        // Collect the data for each view in an array
+        if ('views' in this.config)
+        {
+            this.views = this.config.views;
+
+            for (const view of this.views)
+                view.rawData = deepCloneArray(data);
```
larsxschneider: In context of Git the function name `deepCloneArray` might be confusing. Maybe `deepCopyArray`?

docs/assets/js/_partials/chart-recommended-git-versions.js
```diff
@@ -0,0 +1,118 @@
...
+            ],
+        };
+
+        super(config, canvas, actionBar);
+    }
+
+    render()
+    {
+        const gitVersionsDatabase = new GitVersionsDatabase();
+        const dataURL = '{{ site.dataURL }}/git-versions-new.tsv';
```
larsxschneider: Should we rename the `git-versions-new.tsv` to git-versions.tsv' in this release? Or maybe `git-versions-v2.tsv` to avoid trouble?

docs/users-git-versions.html
```diff
@@ -5,7 +5,38 @@
 ---
 
 <div class="chart-placeholder">
-	<h3>Git Versions</h3>
+	<h3>Recommended Git Versions</h3>
+	<canvas data-type="recommended-git-versions"></canvas>
+	<div class="info-box">
+		<p>
+			Shows how many users connected with recommended, outdated, and vulnerable Git clients.
```
larsxschneider: I am not sure ATM... does it work for SSH? It sure does for HTTPS.

In addition, maybe we should mention that the number are only correct for command line Git clients and UI's that use the command line Git clients under the hood (e.g. SourceTree). GitHub Desktop, SmartGit, etc might have a different name in the logs.

docs/assets/js/_partials/utils.js
```diff
@@ -2,3 +2,8 @@ function deepCloneArray(array)
 {
     return jQuery.extend(true, [], array);
 }
+
+function deepCloneObject(object)
```
larsxschneider: Again, maybe `deepCopy...`?

docs/assets/js/_partials/chart-recommended-git-versions.js
```diff
@@ -9,7 +9,9 @@ class RecommendedGitVersionsChart extends AreaChart
                 vulnerable: 'red',
                 outdated: 'yellow',
                 recommended: 'green',
+                unknown: 'grey',
```
larsxschneider: Some libGit clients report libGit versions numbers that can easily be confused with real Git versions numbers. Would they count here?

docs/tools/update-git-versions.json.sh
pluehne: True, but this script is not meant to be invoked by anyone but us maintainers. As this script just helps keep the release information up-to-date, the vulnerabilities still need to be added manually. Hence, I guess it’s safe to not do thorough checking in this script :slightly_smiling_face:.

docs/tools/update-git-versions.json.sh
pluehne: Sure, we could do that. Given what I said above (the script is just for maintainers), I didn’t put too much priority on polishing the script. Do you think I should improve this for this pull request?

docs/assets/js/_partials/semantic-versioning.js
pluehne: Sure, I’ll address this :+1:!

docs/assets/js/_partials/git-versions-database.js
pluehne: I thought of this as well. However, I think that a front-end setting would be too much work to implement at this stage (after all, this pull request already is gigantic). An easy option might be to add a configuration option to the dashboard front-end to explicitly opt out of the online updates. I’m not sure if anyone would be interested in such an option though. But I might be misjudging the situation, and maybe, some folks are legally restricted in what they are allowed to fetch from online sources within their company’s networks?

docs/assets/js/_partials/chart-area.js
pluehne: Yep. I guess the Chart.js developers wanted to kill as many birds with one stone as possible :smile:.

docs/assets/js/_partials/chart-area.js
pluehne: Well, shallow vs. deep cloning are standard programming terms I guess. I’d keep it like this, because I believe that everybody who’ll touch our source code will understand what’s meant here.

pluehne: See [my other comment below](https://github.com/Autodesk/hubble/pull/174#discussion_r238363234) :point_down:.

pluehne:
@larsxschneider: Thanks for your thorough and constructive review! I’ll address your points as discussed above.

docs/tools/update-git-versions.json.sh
larsxschneider: Nope, it's fine - it was just a nit 👍

docs/assets/js/_partials/git-versions-database.js
larsxschneider: OK. If you leave it as is, then I would make this "loads data from github.com" very explicit in the readme.md. AFAIK we don't load any other resource from the web, right?

pluehne: @larsxschneider: My suggestion would be to write a new (informational, not legally binding) disclaimer section explaining what kinds of data Hubble records from the appliance and fetches from the web. I’d open a separate pull request for that stuff, because it’s not really related to the idea of tracking vulnerable Git versions in the first place.

pluehne:
@larsxschneider: I was so busy in the last weeks, but let’s try and get this merged in the next days.

What do you think are the blockers here? I’d prefer to merge a minimal version of this soon, and we can still address the other issues later as necessary.

pluehne:
@larsxschneider: If you don’t object, I’ll merge this pull request tomorrow. Let’s open separate issues and pull requests for the open ends.

PR #193, opened 2019-10-01
count number of successful connections per protocol

larsxschneider:


pluehne:
@larsxschneider: Thanks!

Issue #179, opened 2019-04-29
Skip Hubble updater execution if GHES config/maintenance is running


larsxschneider:

pluehne:
You have a good point here :slightly_smiling_face:. Is there any way we can detect whether GitHub Enterprise is in maintenance mode?

lbonanomi:
@pluehne: I apologize if this is out of turn, but I can't leave a question unanswered.

`https://$HOSTNAME:8443/setup/api/maintenance?api_key=$CONSOLE_PASSWORD` will return JSON with current maintenance status.

Best,
A Fan

pluehne:
@lbonanomi: No need to apologize, you’re providing *very* useful information! Thanks for letting us know about this API endpoint :+1:. I didn’t know about it yet :slightly_smiling_face:.

PR #195, opened 2019-10-24
Best-effort attempt to check for remote maintenance

lbonanomi:
Per https://github.com/Autodesk/hubble/issues/179 this change attempts to poll a remote instance to check its maintenance status and abort `update-stats.py` if the instance is currently under-maintenance.

Checks are executed with the (batteries-included) http.client and json libraries instead of Requests to enhance portability

### DIFF
--- a/updater/config.py.example
+++ b/updater/config.py.example
@@ -11,6 +11,7 @@ configuration = \
 		"enabled" : False,
 		"sshKey"  : "/path/to/key",
 		"gheHost" : "host",
+		"consolePassword" : "Admin Console Password"
 	},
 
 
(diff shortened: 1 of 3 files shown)


lbonanomi:
@larsxschneider I agree that `ghe-maintenance --query` is simpler, it just seemed prudent to knock on the proverbial door before trying to log-in. 

If you think calling on the maintenance API is the wrong thing to do may I try again with the CLI tools?

larsxschneider:
I would prefer the CLI tools as we already have the SSH key to access the console nicely. Calling the maint API is not wrong, it is just that we would need to store additional credentials to access the API.

lbonanomi:
Closing this PR to pivot the test-for-maintenance model, I'll be back after the weekend with a CLI-based check.

PR #212, opened 2020-10-02
Add a view for failed webhooks

stoe:
## Description

Add a view to display failed webhooks.
This can help to identify webhooks that are "clogging up" the `hookshots` queue and could safely be disabled, e.g. using `ghe-webhook-manage -d 1` on the management console.

```sh
$ ghe-webhook-manage 
Usage: ghe-webhook-manage [options] [hook_id]

Manage webhook status and locate where a hook is configured


OPTIONS:
 -h              Show this message.
 -s              Lookup a webhook's information (target, url, active status)
 -e              Enable a webhook
 -d              Disable a webhook
```

## Demo

1. Open Hubble
2. Go to Housekeeping > Failed Webhooks: `/housekeeping-failed-webhooks`

## Screenshots

hubble failed webhooks

### DIFF
--- a/docs/housekeeping-failed-webhooks.html
+++ b/docs/housekeeping-failed-webhooks.html
@@ -5,7 +5,46 @@ permalink: /housekeeping-failed-webhooks
 ---
 
 <div class="chart-placeholder">
-  <table data-url="{{ site.dataURL }}/failed-webhooks.tsv" data-type="table"></table>
+	<h3>Failed Webhooks</h3>
+	<canvas
+		data-url="{{ site.dataURL }}/failed-webhooks.tsv"
+		data-type="history"
+		data-config='{
+			"views":
+			[
+				{
+					"label": "2 m",
+					"tooltip": "Show the last 2 months",
+					"aggregate": false,
+					"slice": [0, 61],
+					"default": true
+				},
+				{
+					"label": "2 y",
+					"tooltip": "Show the last 2 years",
+					"aggregate":
+					{
+						"period": "week",
+						"method": "sum"
+					},
+					"slice": [0, 106]
+				},
+				{
+					"label": "all",
+					"tooltip": "Show all data",
+					"aggregate":
+					{
+						"period": "week",
+						"method": "sum"
+					}
+				}
+			]
+    }'></canvas>
+	<div class="info-box"></div>
+</div>
+
+<div class="chart-placeholder">
+  <table data-url="{{ site.dataURL }}/failed-webhooks-detailed.tsv" data-type="table"></table>
 
 	<div class="info-box">
 		<p>
--- a/docs/housekeeping-failed-webhooks.html
+++ b/docs/housekeeping-failed-webhooks.html
@@ -1,0 +1,23 @@
+---
+layout: default
+title: Failed Webhooks
+permalink: /housekeeping-failed-webhooks
+---
+
+<div class="chart-placeholder">
+  <table data-url="{{ site.dataURL }}/failed-webhooks.tsv" data-type="table"></table>
+
+	<div class="info-box">
+		<p>
+			<ul>
+        <li><code>business</code> a <a href="https://docs.github.com/enterprise-server/admin/user-management/managing-global-webhooks">global webhook</a></li>
+        <li><code>integration</code> a <a href="https://docs.github.com/enterprise-server/developers/apps/getting-started-with-apps">GitHub App</a> webhook URL</li>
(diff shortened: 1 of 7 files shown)


stoe:
### Updated Screenshot
for 321a36a

image

stoe:
@shegox

updater/scripts/failed-webhooks.sh
```diff
@@ -0,0 +1,26 @@
...
+  sort |
+  uniq -ic |
+  sort -rn |
+  awk -F [ '{printf("%s,%s\n", $2,$1)}' |
+  awk -F, '{
+    gsub(/(\[|\]|\")/, "");
+    gsub(/-[0-9]+/, "", $3);
+    gsub(/ /, "", $5);
+    printf("%s\t%s\t%s\t%s\t%s\n",$1,$3,$2,$4,$5)
+  }'
```
larsxschneider, line 26:
```suggestion
   perl -pne 's/([0-9]+)\s(.*)/\2\t\1/'
```
I think we can combine them like this

larsxschneider (changes requested):
Very nice! Let's fix some minor nit and ship this! Thank you for implementing this 🙇

docs/housekeeping-failed-webhooks.html
```diff
@@ -0,0 +1,62 @@
...
+    }'></canvas>
+	<div class="info-box"></div>
+</div>
+
+<div class="chart-placeholder">
+  <table data-url="{{ site.dataURL }}/failed-webhooks-detailed.tsv" data-type="table"></table>
+
+	<div class="info-box">
+		<p>
+			<ul>
```
larsxschneider, line 51: We might want to add some text here explaining why webhooks could fail (e.g. because the server is down, DNS resolution is not possible, webhook server error, ...) ... but we can work on this in a subsequent PR I think

stoe: Isn't that already resolved by the `message` column in the table, @larsxschneider?

larsxschneider: Very briefly, yes. We usually try to give admins some additional context in that column. E.g. a useful thing would be to teach team about `ghe-webhook-logs` to investigate a webhook error furthe. In the table then will see the failing hook id and then they could run to learn more:
```
ghe-webhook-logs --failures --hook-id <hook-id> 
```

updater/reports/ReportFailedWebhooks.py
```diff
@@ -0,0 +1,14 @@
...
+# Report how many webhooks failed
+class ReportFailedWebhooks(ReportDaily):
+	def name(self):
+		return "failed-webhooks"
+
+	def updateDailyData(self):
+		self.detailedHeader, self.detailedData = self.parseData(
+			self.executeScript(self.scriptPath("failed-webhooks.sh")))
+		self.header = ["date", "failed webhooks"]
+		self.data.append([str(self.yesterday()), len(self.detailedData)])
```
larsxschneider, line 12:
```suggestion
		self.data.append(
			[
				str(self.yesterday()),
				sum(map(lambda x: int(x[4] if len(x) > 3 else 0), newData)),
			])
		self.detailedData = newData[:25]
```
After looking at this with fresh eyes, I think here is a little bug.

For the chart we want to count the number of failures *for each* failed webhook (IIRC the count column is `4`). That's why we need the sum function.

For the detailed view (the table). We just want to write out the results. Since on a large instance many many hooks could fail we limit the length of the list to 25 here.

Disclaimler: This is untested and just "code in the GitHub review window" 😨

larsxschneider (changes requested):

updater/reports/ReportFailedWebhooks.py
```diff
@@ -0,0 +1,14 @@
+from .ReportDaily import *
+
+# Report how many webhooks failed
+class ReportFailedWebhooks(ReportDaily):
+	def name(self):
+		return "failed-webhooks"
+
+	def updateDailyData(self):
+		self.detailedHeader, self.detailedData = self.parseData(
```
larsxschneider, line 9:
```suggestion
		self.detailedHeader, newData = self.parseData(
```
This is necessary for the change below

larsxschneider (commented):
Woohoo! Thank you @stoe 🙇

pluehne:
Great job, @stoe and @larsxschneider, I like this addition very much!

docs/housekeeping-failed-webhooks.html
stoe: Will open a new PR for this soon™

mlbright:
Nice work @stoe and @larsxschneider , we have already used this to dramatically cut the number of failed webhooks. Thank you!

PR #213, opened 2020-10-07
add option to run the Hubble updater with Docker

larsxschneider:
/cc @kammerjaeger

### DIFF
--- a/updater/Dockerfile
+++ b/updater/Dockerfile
@@ -13,7 +13,7 @@ WORKDIR $APP_HOME
 # Copy over our application code
 ADD reports $APP_HOME/reports
 ADD scripts $APP_HOME/scripts
-ADD ./check-schema.py ./helpers.py ./schema.py ./update-stats.py $APP_HOME/
+ADD ./*.py $APP_HOME/
 
 # Install dependencies
 RUN apk add --no-cache git openssh-client
--- a/updater/Dockerfile
+++ b/updater/Dockerfile
@@ -1,0 +1,22 @@
+# Run the Hubble updater with Docker
+
+FROM python:alpine
+
+# Set environment variables
+ENV APP_HOME /hubble-updater
+
+# Create a directory for our application
+# and set it as the working directory
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+# Copy over our application code
+ADD reports $APP_HOME/reports
+ADD scripts $APP_HOME/scripts
+ADD ./check-schema.py ./helpers.py ./schema.py ./update-stats.py $APP_HOME/
+
+# Install dependencies
+RUN apk add --no-cache git openssh-client
+
+# Run the app
+CMD ["python3", "update-stats.py"]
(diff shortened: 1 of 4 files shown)


updater/README.md
```diff
@@ -18,6 +18,17 @@ The updater may be run in two modes:
 1. Create a config file in `updater/config.py` in your local clone (based on [`updater/config.py.example`](config.py.example))
 1. Configure a service or cronjob that executes `python3 update-stats.py` once per day.
 
+Alternatively you can set it up via Docker:
```
pluehne, line 21:
```suggestion
Alternatively, you can set it up via Docker:
```

updater/README.md
```diff
@@ -18,6 +18,17 @@ The updater may be run in two modes:
...
+1. Build the Docker image
+    ```sh
+    $ docker build . --tag hubble-updater
+    ```
+1. Create a config file in `updater/config.py` (based on [`updater/config.py.example`](config.py.example))
+1. Configure a service or cronjob that executes the docker container once a day (replace `/path/to/config.py` and `/path/to/key` with your paths):
```
pluehne, line 27:
```suggestion
1. Configure a service or cron job that executes the Docker container once a day (replace `/path/to/config.py` and `/path/to/key` with your paths):
```

updater/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# Create a directory for our application
+# and set it as the working directory
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+
+# Copy over our application code
+ADD reports $APP_HOME/reports
+ADD scripts $APP_HOME/scripts
+ADD ./check-schema.py ./helpers.py ./schema.py ./update-stats.py $APP_HOME/
```
pluehne, line 16: Should we use a wildcard instead?
```suggestion
ADD ./*.py $APP_HOME/
```
I’m just wondering whether that would make it more future-proof (we may add or remove files later on, in which case this Dockerfile would break if we forgot to update it accordingly).

pluehne (approved):
Looks good to me (just check my minor suggestions)!

kammerjaeger:
Would it not make more sense to have the cron in the docker container? Then no external interaction is needed.
Another thing, is it planed that there will be an official container uploaded per release? That would make it even easier to deploy it.

updater/README.md
pluehne: Also, I’d recommend a `systemd` service over a cron job.

larsxschneider:
> Would it not make more sense to have the cron in the docker container? Then no external interaction is needed.

I'll look into that 👍 

> Another thing, is it planed that there will be an official container uploaded per release? That would make it even easier to deploy it.

Yeah, we could add a GitHub action which pushes a new docker release on every change. But I would like to do that in a separate PR.

updater/README.md
larsxschneider: Hmm. Unfortunately, Alpine doesn't use systemd ... therefore I think it is easier to just use cron. OK?

pluehne: Absolutely, didn’t know about that :+1:.

PR #222, opened 2020-12-17
warn user about incorrect traffic and versions numbers with active replicas

larsxschneider:
`/var/log/github-audit.log` is used to query traffic and version info.
These log files are written on the node that serves the traffic and
Hubble only queries the log file on the primary. Therefore, in a configuration
with active replicas we are not reporting the right numbers.

Warn the user about it until we fix this.


pluehne (approved):
Good suggestion, let’s notify users about that.

PR #228, opened 2021-04-28
Fixing GHES 3.0 Compatibility

svasek:
GitHub uses containerization of more services on version 3.0 as opposed to version 2.22. That means containers use the journald log driver and logs are available via journalctl, rather than the usual log location on disk as they used to be in the past. This may be affecting 3rd party's tool log parsing logic.


stoe (commented):
Small suggestions for backwards compatibility.

svasek:
Replaced by #235

PR #244, opened 2022-03-03
Separate command arguments from name

pluehne:
In order to execute commands, `subprocess.Popen` expects a list of strings, where the first one represents the name of the command and the remaining strings are the command’s arguments. However, the `-s` option was included in the name of the binary, which isn’t supported in newer versions of Python anymore.


pluehne:
@svasek: I took this contribution of yours out of #236 because we needed to make unrelated changes. This one is unproblematic and quick to review, so I turned it into a separate pull request.

mariusknaust (approved):

PR #245, opened 2022-03-03
Refactor functions for server calls

pluehne:
The function `executeScript` used to have a double role, as it was in charge of executing both Unix commands on the server (locally or by using a local connection) but also Bash scripts.

This was confusing per se, but had a few side effects that might have been unexpected to developers looking at the code.

First of all, the optional `stdin` argument was completely ignored when executing Bash scripts on the server. The reason for that is that `stdin` was used to pass the script’s content to the server (using the `-s` option provided by Bash). In a sense, the interface of this function was misleading as a result of this.

Also, when using `executeScript` to run regular Unix commands, unnecessary file look-ups were made.

To address these issues, this replaces the `executeScript` method with two new ones: `executeCommandOnServer` (which executes regular Unix commands) and `executeBashScriptOnServer` (which executes Bash scripts and currently does not accept stdin because of the limitation mentioned above).

In addition to that, other similar functions are renamed to `executeDatabaseQueryOnServer` and `executeRubyScriptOnServer` for consistency and clarity.

Finally, scripts are now read from the `scripts/` directory by default to simplify calls of `executeScriptOnServer`.

### DIFF
--- a/updater/reports/Report.py
+++ b/updater/reports/Report.py
@@ -89,7 +89,7 @@ class Report(object):
 
 		return stdout
 
-	def executeGHEConsole(self, script):
+	def executeRubyScriptOnServer(self, script):
 		# Escape the Ruby script, as it is going to be sent as a string on the command line
 		script = script.replace('\\t', '\\\\t').replace('\\n', '\\\\n')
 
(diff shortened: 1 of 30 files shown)


mariusknaust (approved):
Looks good with the fix of #247 in mind.

PR #235, opened 2021-08-26
Add GHES 3.x compatibility

stoe:
- Use GHES 3.x syntax for Git requests
- Use GHES 3.x syntax for Git protocol usage
- Use GHES 3.x syntax for failed webhooks

Also addresses review comments in https://github.com/Autodesk/hubble/pull/228

PR #246, opened 2022-03-03
Adjust log files for GitHub Enterprise Server 3.x

pluehne:
Some log file paths have changed between GitHub Enterprise Server 2.x and 3.x, and this commit adjusts for that.


pluehne:
@stoe, @svasek: Thanks for your contributions to make Hubble compatible with GitHub Enterprise Server 3.x, and sorry also for not being able to work on this earlier!

While looking into this, I realized two things: First, GitHub Enterprise Server 2.x is out of support by GitHub, so I decided to remove support for these versions from Hubble as well, which allows us to remove all those version switches. Second, the way Hubble executed commands and scripts on the GitHub server was a little problematic, which is why we decided to refactor those parts while adding support for GitHub Enterprise Server 3.x in a very similar way to what you worked on in #235 and #236.

As a result of those changes, it would have been a lot of work to make your pull requests compatible, so I decided to create this pull request to supersede them (I’ll close #235 and #236 when this one is merged). I hope that’s fine by you. I made sure that you are listed as contributors in the commits so you will receive credit for your contributions :slightly_smiling_face:. Thanks again!

If you have some time, could you run this version in your environments and let us know if it’s working as expected? We did some testing on our end but would like to get some feedback from actual users before merging this, as the changes are a little more invasive than we first had in mind.

stoe:
> As a result of those changes, it would have been a lot of work to make your pull requests compatible, so I decided to create this pull request to supersede them (I’ll close #235 and #236 when this one is merged). I hope that’s fine by you. I made sure that you are listed as contributors in the commits so you will receive credit for your contributions 🙂. Thanks again!

Awesome @pluehne.
I'm more than happy if #235 gets closed, even before this  one here is merged.

pluehne:
While testing this, we discovered another issue. @stoe, @svasek: In case you want to test these changes, please use the state in #247, as that fix appears to be necessary to get reliable results now.

mariusknaust (approved):

